PR #1, opened 2025-01-15
2024 Vision Subsystem added

kujo27:
Only thing that changed was a different update() method in getPhotonVisionEstimatedPose().


JohnGilb (approved):
Good to start, let's try it out on 2024

PR #8, opened 2025-01-18
Algae Collection Skeleton

AlanYuan99:
Created motors for algae intake and output.

No questions.

No tests.


src/main/java/competition/electrical_contract/ElectricalContract.java
```diff
@@ -17,6 +17,8 @@ public abstract class ElectricalContract implements XSwerveDriveElectricalContra
 
     public abstract CANMotorControllerInfo getSteeringMotor(SwerveInstance swerveInstance);
 
+    public abstract CANMotorControllerInfo getAlgaeCollectionMotor();
```
aschokking, line 20: ⭐ ⭐ ⭐ defining this here is great, but this only promises that later on in the year specific contracts you'll provide the actual motor information (what kind of motor is it, what channel is it running on), we need to do that or the code won't compile.

aschokking, line 20: ⭐ ⭐ ⭐ Another thing that would be handy to add here is a `boolean isAlgaeCollectionReady()` method that we can use to tell the robot is the motor actually on the robot yet (which it isn't currently). See above for an example of us doing that already for the drive.

Add that had have the 2025 contract say `false` that it's not ready yet.

src/main/java/competition/subsystems/AlgaeCollectionSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+@Singleton
+public class AlgaeCollectionSubsystem extends BaseSubsystem {
+    public final XCANMotorController motor;
+
+    public DoubleProperty intakePower;
+    public DoubleProperty outputPower;
+
+    @Inject
+    public AlgaeCollectionSubsystem(XCANMotorController.XCANMotorControllerFactory xcanMotorControllerFactory, ElectricalContract electricalContract, PropertyFactory propertyFactory) {
+        this.motor = xcanMotorControllerFactory.create(electricalContract.getAlgaeCollectionMotor(), getPrefix(), "AlgaeMotor");
```
aschokking, line 21: ⭐ ⭐ ⭐ here you can check if `isAlgaeCollectoReady()` and if it's not don't create the motor (just leave it null).

src/main/java/competition/subsystems/AlgaeCollectionSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+
+        this.intakePower = propertyFactory.createPersistentProperty("Intake Power", .1);
+        this.outputPower = propertyFactory.createPersistentProperty("Output Power", .1);
+    }
+
+    public void intake() {
+        motor.setPower(intakePower.get());
```
aschokking, line 28: ⭐ ⭐ ⭐ here you can check if `isAlgaeCollectoReady()` and if it's not don't set the power

src/main/java/competition/subsystems/AlgaeCollectionSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public void output() {
+        motor.setPower(outputPower.get());
```
aschokking, line 32: ⭐ ⭐ ⭐ here you can check if `isAlgaeCollectoReady()` and if it's not don't set the power

src/main/java/competition/subsystems/AlgaeCollectionSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public void stop() {
+        motor.setPower(0);
```
aschokking, line 36: ⭐ ⭐ ⭐ here you can check if `isAlgaeCollectoReady()` and if it's not don't set the power

src/main/java/competition/subsystems/collection/IntakeCommand.java
aschokking, line 1: ⭐ Since we might have many things on the robot called 'Intake', it would help disambiguate what this command does by being more explicit: `AlgaeCollectionIntakeCommand` for example. Please update the 3 commands to this naming pattern

src/main/java/competition/subsystems/AlgaeCollectionSubsystem.java
```diff
@@ -0,0 +1,39 @@
+package competition.subsystems;
```
aschokking, line 1: ⭐ ⭐ ⭐ this subsystem file should be nested down 1 level in a folder (like you have the commands). But let's make that folder name more explicitly `algae_collection` rather than just `collection` (because we'll also have Coral collection later).

aschokking (changes requested):
Great start! A bunch of things we forgot to tell you about initially to tweak

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -31,6 +31,18 @@ public boolean isDriveReady() {
         return true;
     }
 
+    public boolean isAlgaeCollectionReady() { return false;}
```
aschokking, line 34: ⭐ nit: space after `; `

src/main/java/competition/subsystems/algae_collection/AlgaeCollectionSubsystem.java
```diff
@@ -0,0 +1,53 @@
...
+
+@Singleton
+public class AlgaeCollectionSubsystem extends BaseSubsystem {
+    public final XCANMotorController motor;
+
+    public DoubleProperty intakePower;
+    public DoubleProperty outputPower;
+
+    @Inject
+    public AlgaeCollectionSubsystem(XCANMotorController.XCANMotorControllerFactory xcanMotorControllerFactory, Contract2025 contract2025, PropertyFactory propertyFactory) {
```
aschokking, line 22: ⭐ ⭐ ⭐ Ask for `ElectricalContract` here, not a specific year, otherwise we can't switch on different robots

aschokking (changes requested):
Really close! A couple more small changes

aschokking (approved):
🎉

PR #14, opened 2025-01-18
Elevator subsystem skeleton creation

junyu101:
Creating elevator subsystem skeleton

Asana task URL: https://app.asana.com/0/1208731987242938/1209174882393849/f

# Whats changing?
Basic Elevator Commands inlcluding rise, lower, stop, along with basic power setting to elevator motors, also added abstract methods for motor device info in electrical contract, and implemented them in competition contract 2025.

Does the Elevator subsystem have to extend a BaseSetpointSubsystem rather than a BaseSubsystem
Also are the motors supposed to be XCANTalon's or is it the XCANMotorControllers
What the heck is motion magic and do i need to use that for elevator power logic in the future?

No tests, just a basic skeleton
-----

PR feedback legend

 

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -36,6 +37,23 @@ public boolean areCanCodersReady() {
...
 
+    @Override
+    public boolean isElevatorReady() {
+        return false;
+    }
+
+    @Override
+    public CANMotorControllerInfo getElevatorMaster() {
+        return new CANMotorControllerInfo(
+                "Master Motor", MotorControllerType.TalonFx, new CANBusId("99"), 99, new CANMotorControllerOutputConfig());
```
aschokking, line 48: ⭐ ⭐ "Master motor" is kind of ambiguous, what if we have multiple motor pairs on the robot? What is this motor controlling?

aschokking, line 48: We don't know yet if it will be a talon or sparkmax so having anything as a placeholder is fine, we can change it later

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,79 @@
...
+import xbot.common.properties.PropertyFactory;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class ElevatorSubsystem extends BaseSubsystem {
+
+    public enum ElevatorGoal{
+        //common TargetHeights
```
aschokking, line 20: ⭐ you can probably define these now right?

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,79 @@
...
+import xbot.common.controls.actuators.XCANTalon;
+import xbot.common.injection.electrical_contract.CANTalonInfo;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.PropertyFactory;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class ElevatorSubsystem extends BaseSubsystem {
```
aschokking, line 17: ⭐ Yeah I think it makes sense for this to be a base setpoint subsystem, but let's talk about why

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public enum ElevatorGoal{
+        //common TargetHeights
+    }
+    double defaultElevatorPower;
```
aschokking, line 22: ❓ what's the point of this?

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,79 @@
...
+    final DoubleProperty elevatorPower;
```
aschokking, line 23: ⭐ ⭐ Making the elevator go up vs down usually needs VERY different powers. And maybe we just have a gamepad control it with the joystick in the short-run?

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,79 @@
...
+    //assuming we'll have a master and follower motors
+    public XCANMotorController master;
+    public XCANMotorController follower;
+
+
+    @Inject
+    public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFactory, PropertyFactory pf, ElectricalContract contract){
+
+        defaultElevatorPower = 0.5;
+        elevatorPower = pf.createPersistentProperty("Standard Power", 0.5);
```
aschokking, line 39: ⭐ it's a good idea to make this variable name more closely aligned with the string that shows up in the dashboard

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,111 @@
...
+            this.followerMotor = motorFactory.create(contract.getElevatorFollower(), this.getPrefix(), "Elevator Motor");
+        }
+    }
+
+    //will implement logic later
+    @Override
+    public void setPower(Double power) {
+        if (contract.isElevatorReady()) {
+            masterMotor.setPower(power);
+            followerMotor.setPower(power);
```
JohnGilb, line 60: just FYI, even though we're only using one motor, if we were using two the follower would "follow automatically; no need to set powers here.

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,111 @@
...
+import xbot.common.command.BaseSetpointSubsystem;
+import xbot.common.controls.actuators.XCANMotorController;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.PropertyFactory;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class ElevatorSubsystem extends BaseSetpointSubsystem<Double> implements DataFrameRefreshable {
```
JohnGilb, line 14: Let's use `Distance` rather than `Double` as a maintainer.

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,111 @@
...
+    //will implement logic later
+    @Override
+    public void setPower(Double power) {
+        if (contract.isElevatorReady()) {
+            masterMotor.setPower(power);
+            followerMotor.setPower(power);
+        }
+    }
+
+    public void rise(){
```
JohnGilb, line 64: recommend renaming to "raise".
Raise & Lower are a semantic pair.
Rise & Fall are a semantic pair.

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    protected boolean areTwoTargetsEquivalent(Double target1, Double target2) {
+        return BaseSetpointSubsystem.areTwoDoublesEquivalent(target1, target2, 1);
+    }
+
+    @Override
+    public void refreshDataFrame() {
+        //to be implemented later
```
JohnGilb, line 107: just do `motor.refreshDataFrame()` for any motor(s) you have.

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
```
JohnGilb, line 109: also add a `periodic()` where you call `motor.periodic()`

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -0,0 +1,35 @@
+package competition.subsystems.elevator.commands;
+
+import competition.subsystems.elevator.ElevatorSubsystem;
+import xbot.common.command.BaseCommand;
+import xbot.common.command.BaseSubsystem;
+import xbot.common.properties.PropertyFactory;
+
+import javax.inject.Inject;
+
+public class ElevatorMaintainerCommand extends BaseCommand {
```
JohnGilb, line 10: this needs to extend `BaseMaintainerCommand`

src/main/java/competition/subsystems/elevator/commands/SetElevatorTargetHeightCommand.java
```diff
@@ -0,0 +1,33 @@
+package competition.subsystems.elevator.commands;
+
+import competition.subsystems.elevator.ElevatorSubsystem;
+import xbot.common.command.BaseCommand;
+
+import javax.inject.Inject;
+
+public class SetElevatorTargetHeightCommand extends BaseCommand {
```
JohnGilb, line 8: this needs to extend `BaseSetpointCommand`

JohnGilb (changes requested):

PR #20, opened 2025-01-21
Swerve Fixes for 2023/2024/2025 Chassis

JohnGilb:
# Why are we doing this?
- Fixing various bugs found when deploying 2025 code to all recent chassis
Asana task URL:

# What's changing?
-Bugfixes in contracts
-Temporary disable of vision system for 2024 until PhotonVision version mismatch addressed

# How this was tested
- [x] tested on robot

### DIFF
--- a/src/main/java/competition/Robot.java
+++ b/src/main/java/competition/Robot.java
@@ -40,9 +40,9 @@ public class Robot extends BaseRobot {
     protected BaseRobotComponent createDaggerComponent() {
         if (BaseRobot.isReal()) {
             // Initialize the contract to use if this is a fresh robot. Assume competition since that's the safest.
-            if (!Preferences.containsKey("ContractToUse")) {
+            /*if (!Preferences.containsKey("ContractToUse")) {
                 Preferences.setString("ContractToUse", "Competition");
-            }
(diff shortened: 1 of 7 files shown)


src/main/java/competition/Robot.java
```diff
@@ -57,6 +53,12 @@ protected BaseRobotComponent createDaggerComponent() {
                     System.out.println("Using Robox contract");
                     return DaggerRoboxComponent.create();
                 default:
+                    // Moved setting the default contract to here; there was some bug where
+                    // other chassis were being set to "Competition" when they shouldn't have been.
+                    // Root cause is unknown, so far now only setting this after checking for any other value.
+                    if (!Preferences.containsKey("ContractToUse")) {
```
stephenjust, line 59: Why do we need to set this at all? Network Tables will use the default on get if it's missing.

stephenjust (approved):

src/main/java/competition/Robot.java
JohnGilb: Creating the key will allow users to change the value in other tools like Glass/Shuffleboard/etc in case the default is wrong.

PR #26, opened 2025-01-24
Terminal approach + AprilTagVisionSubsystemExtended

kujo27:
# Why are we doing this?
Allows us to drive to the nearest reef face to score using SwerveSimpleTrajectoryCommand and apriltags. Also, created a subsystem to be able to simulate in advantagescope
Asana task URL:
https://app.asana.com/0/1208731987242938/1209104793195512/f
# Whats changing?
New files are added: DriveToNearestReefFaceCommand, AlignToReefWithAprilTagCommand, DriveToNearestReefFaceCommandGroup, MockAprilTagVisionSubsytem
Need to push some changes to SCL for there to be no errors ( new move() in BaseSwerveDriveSubsystem, and make inputs[] public in AprilTagVisionSubsystem )
# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 

src/main/java/competition/subsystems/vision/MockAprilTagVisionSubsystem.java
aschokking, line 1: ⭐ ⭐ ⭐ please rename this to not be mock

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
```diff
@@ -0,0 +1,111 @@
...
+    PoseSubsystem pose;
+
+    public static Pose2d BlueCloseReefFace = new Pose2d(3.158, 4.026, Rotation2d.fromDegrees(0));
+    public static Pose2d BlueCloseRightReefFace = new Pose2d(3.829, 2.880, Rotation2d.fromDegrees(60));
+    public static Pose2d BlueCloseLeftReefFace = new Pose2d(3.829, 5.180, Rotation2d.fromDegrees(-60));
+    public static Pose2d BlueBackLeftReefFace = new Pose2d(5.150, 5.180, Rotation2d.fromDegrees(-120));
+    public static Pose2d BlueBackReefFace = new Pose2d(5.821, 5.821 , Rotation2d.fromDegrees(-180));
+    public static Pose2d BlueBackRightReefFace = new Pose2d(5.150, 2.880, Rotation2d.fromDegrees(120));
+
+    public double cameraOffset = 0.5;
```
aschokking, line 29: ⭐ ⭐  what is this? what are the units? Use a Distance class instead maybe?

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
```diff
@@ -0,0 +1,111 @@
...
+    public void initialize() {
+        drive.getPositionalPid().reset();
+    }
+
+    @Override
+    public void execute() {
+        Translation2d driveValues = getXYPowersAlignToAprilTag();
+
+        double omega = headingModule.calculateHeadingPower(getClosestReefFacePose().getRotation().getDegrees());
+
```
aschokking, line 51: ⭐ should we maybe send to aKit what the closest reef face pose was? Like where we're trying to go?

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
```diff
@@ -0,0 +1,111 @@
...
+        this.aprilTagVisionSubsystem = aprilTagVisionSubsystem;
+        this.drive = drive;
+        this.headingModule = headingModuleFactory.create(drive.getRotateToHeadingPid());
+        this.pose = pose;
+        addRequirements(drive);
+    }
+
+    @Override
+    public void initialize() {
+        drive.getPositionalPid().reset();
```
aschokking, line 43: ⭐ maybe here in init we should log the closest reef face? Is there a reason we shouldn't check it once in init and lock it in vs continously checking for closest in execute?

aschokking: Also in practice we'll probably have a target reef to go to rather than just closest, maybe we should tweak this command later to take that as an input?

aschokking (approved):

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -0,0 +1,73 @@
...
+public class AprilTagVisionSubsystemExtended extends AprilTagVisionSubsystem {
+    @Inject
+    public AprilTagVisionSubsystemExtended(PropertyFactory pf,
+                                           AprilTagFieldLayout fieldLayout, XCameraElectricalContract contract,
+                                           AprilTagVisionIOFactory visionIOFactory) {
+        super(pf, fieldLayout, contract, visionIOFactory);
+    }
+
+    @Override
+    public void periodic() {
```
JohnGilb, line 30: You can omit this whole function if all you do is call `super`

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -0,0 +1,73 @@
...
+        super.periodic();
+    }
+
+    @Override
+    public void refreshDataFrame() {
```
JohnGilb, line 35: same as above comment

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -0,0 +1,73 @@
...
+        return targetObservation.fiducialId() == targetAprilTagID;
+    }
+
+    public int getTargetAprilTagID(Pose2d targetReefFacePose) {
+        HashMap<Pose2d, Integer> hashMap = new HashMap<>();
+
+        // Note: flipped april tag IDs across the y-midpoint of the field
+        // map Red Alliance sided April Tags
+        if (DriverStation.getAlliance().orElse(DriverStation.Alliance.Blue) == DriverStation.Alliance.Red) {
+            hashMap.put(PoseSubsystem.convertBlueToRedIfNeeded(Landmarks.BlueCloseLeftAlgae), 8);
```
JohnGilb, line 56: create and populate the hashmap in the constructor (so this only happens one time) with both red and blue. Then only get from the list in this function

JohnGilb (changes requested):

PR #28, opened 2025-01-24
Arm Simulation

aschokking:
# Why are we doing this?

To simulate the arm mechanism

Asana task URL:

# Whats changing?

Modeled off the WPIlib arm example project, using all it's default values for things. This needs more work but is a start.

## Future work

We need to model the encoder rotations and angles more specifically

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

Not tested at all yet
-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/simulation/arm/ArmSimConstants.java
+++ b/src/main/java/competition/simulation/arm/ArmSimConstants.java
@@ -1,0 +1,22 @@
+package competition.simulation.arm;
+
+import static edu.wpi.first.units.Units.Degrees;
+import static edu.wpi.first.units.Units.Kilogram;
+import static edu.wpi.first.units.Units.Meters;
+
+import edu.wpi.first.units.measure.Angle;
+import edu.wpi.first.units.measure.Distance;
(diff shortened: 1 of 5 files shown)


src/main/java/competition/simulation/arm/ArmSimConstants.java
```diff
@@ -0,0 +1,22 @@
...
+import edu.wpi.first.units.measure.Angle;
+import edu.wpi.first.units.measure.Distance;
+import edu.wpi.first.units.measure.Mass;
+
+public class ArmSimConstants {
+    // TODO: this is all from the wpilib example but doesn't reflect our real math
+    public static final double armReduction = 200;
+    public static final Mass armMass = Kilogram.of(8.0); // Kilograms
+    public static final Distance armLength = Meters.of(0.5);
+    public static final Angle minAngleRads = Degrees.of(-45);
```
aschokking, line 16: Some of this stuff needs to be shared with other parts of the program eventually

PR #32, opened 2025-01-25
Elevator in simulator: read from mock motor 

aschokking:
# Why are we doing this?

Now that we have an ElevatorSubsystem, we can read from it to control the simulated elevator mechanism

Asana task URL:

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

Untested because the elevator can't actually be controlled yet

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/simulation/ElevatorSimulator.java
+++ b/src/main/java/competition/simulation/ElevatorSimulator.java
@@ -1,14 +1,17 @@
 package competition.simulation;
 
 import static edu.wpi.first.units.Units.Meters;
+import static edu.wpi.first.units.Units.Rotations;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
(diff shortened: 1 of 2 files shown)


src/main/java/competition/simulation/ElevatorSimulator.java
```diff
@@ -43,28 +49,19 @@ public ElevatorSimulator(ElevatorMechanism elevatorMechanism) {
     }
 
     public void update() {
-        this.elevatorSim.setInputVoltage(elevatorVoltage);
+        this.elevatorSim.setInputVoltage(this.motor.getPower());
```
JohnGilb, line 52: `getPower` returns values between -1 and 1, but this looks like it wants a voltage.

JohnGilb (changes requested):

PR #34, opened 2025-01-25
Allow using Vision pose as the source of robot pose, and pipe it through to simulator

stephenjust:
# Why are we doing this?

Asana task URL:

# Whats changing?
Added a switch to use vision assisted pose vs odometry only pose.

# How this was tested
- [x] unit tests added
- [x] tested on simulated robot

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/simulation/MapleSimulator.java
+++ b/src/main/java/competition/simulation/MapleSimulator.java
@@ -2,19 +2,13 @@ package competition.simulation;
 
 import competition.simulation.arm.ArmSimulator;
 import competition.subsystems.drive.DriveSubsystem;
-import competition.subsystems.elevator.ElevatorMechanism;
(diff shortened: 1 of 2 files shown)


src/main/java/competition/simulation/MapleSimulator.java
```diff
@@ -87,6 +91,12 @@ protected void updateDriveSimulation() {
                 drive.getRearLeftSwerveModuleSubsystem().getTargetState(),
                 drive.getRearRightSwerveModuleSubsystem().getTargetState()
         });
+        vision.getAllPoseObservations().forEach(observation -> {
```
aschokking, line 94: ⭐ ⭐ ⭐ I _think_ we don't need/want to do this since we're not using this internal poseEstimator for anything

aschokking (approved):

PR #36, opened 2025-01-25
Unit Test Contract

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

made booleans true in unit test

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/electrical_contract/UnitTestContract2025.java
+++ b/src/main/java/competition/electrical_contract/UnitTestContract2025.java
@@ -13,4 +13,9 @@ public class UnitTestContract2025 extends Contract2025 {
 
     public boolean isArmPivotMotorReady() { return true; }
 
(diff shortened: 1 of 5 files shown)


src/main/java/competition/electrical_contract/UnitTestContract2025.java
```diff
@@ -6,4 +6,11 @@ public class UnitTestContract2025 extends Contract2025 {
 
     @Inject
     public UnitTestContract2025() {}
+
```
aschokking, line 9: ⭐ ⭐ ⭐ there are more isReady*() calls you haven't added here (like the elevator)

aschokking (approved):
The ones that are here look good, but some are missing

aschokking (approved):

PR #39, opened 2025-01-25
Gamepad control elevators

junyu101:
# Why are we doing this?
To test if we can allocate power to elevator motors 

Asana task URL: https://app.asana.com/0/1208731987242938/1209231306808573/f

# Whats changing?
Just changes to elevator subsystem/maintainer and some dependencies, along with a small change in SCL to fix voltage and current not displaying in the sim

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -60,13 +60,16 @@ public boolean areCanCodersReady() {
 
     @Override
     public boolean isElevatorReady() {
-        return false; //return true when ready
+        return true; //return true when ready
```
JohnGilb, line 72: this needs to be false until the robot has an elevator.

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -113,12 +115,16 @@ protected boolean areTwoTargetsEquivalent(Distance target1, Distance target2) {
...
-        masterMotor.refreshDataFrame();
+        if(contract.isElevatorReady()){
+            masterMotor.refreshDataFrame();
+        }
     }
 
     public void periodic(){
         masterMotor.periodic();
         aKitLog.record("ElevatorTargetHeight",elevatorTargetHeight);
+        aKitLog.record("ElevatorMotorVoltage", masterMotor.getVoltage());
```
JohnGilb, line 126: Remove these two - these values are inputs, and as such are automatically recorded by AKIT.

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -39,33 +48,49 @@ public void initialize() {
 
     @Override
     public void execute(){
```
JohnGilb, line 50: no need to even have this method if you just call super.execute

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -39,33 +48,49 @@ public void initialize() {
...
+
+        return Math.abs(current.in(Inches)/target.in(Inches));
     }
 
     @Override
     protected double getHumanInput() {
-
-        return 0;
+        return MathUtils.deadband(
+                oi.programmerGamepad.getLeftVector().getY(),
```
JohnGilb, line 87: this should probably be the Operator gamepad (which you may need to make)

JohnGilb (changes requested):

src/main/java/competition/electrical_contract/Contract2025.java
aschokking: it's the unit test contract that wants to mark it true so you can use it in the simulator

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
aschokking: No I think this is supposed to be the 'fundamentals' gamepad (which has been called programmer for some reason this year)

PR #41, opened 2025-01-25
Coral sensor

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

Created code for the coral sensor.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/electrical_contract/Contract2025.java
+++ b/src/main/java/competition/electrical_contract/Contract2025.java
@@ -62,6 +62,13 @@ public class Contract2025 extends ElectricalContract {
                 new CANMotorControllerOutputConfig());
     }
 
+    public boolean isCoralSensorReady() { return false; }
+
+    @Override
+    public DeviceInfo getCoralSensor() {
+        return new DeviceInfo("CoralSensor", 0);
+    }
+
     @Override
     public boolean areCanCodersReady() {
         return true;
--- a/src/main/java/competition/subsystems/coral_scorer/CoralScorerSubsystem.java
+++ b/src/main/java/competition/subsystems/coral_scorer/CoralScorerSubsystem.java
@@ -1,6 +1,7 @@
 package competition.subsystems.coral_scorer;
 
 import competition.electrical_contract.ElectricalContract;
+import xbot.common.advantage.DataFrameRefreshable;
 import xbot.common.command.BaseSubsystem;
 import xbot.common.controls.actuators.XCANMotorController;
 import xbot.common.controls.sensors.XDigitalInput;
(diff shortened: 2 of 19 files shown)


src/main/java/competition/subsystems/coral_scorer/CoralScorerSubsystem.java
```diff
@@ -30,26 +34,39 @@ public CoralScorerSubsystem(XCANMotorController.XCANMotorControllerFactory xcanM
...
     }
-    public void stop(){
+    public void stop() {
         if (motor!=null){
             motor.setPower(0);
         }
     }
-
-
+    public boolean hasCoral() {
```
JohnGilb, line 65: this method needs to:
* If the sensor is not ready, return `false`
* If the sensor is ready, return the value of `get()`

JohnGilb (changes requested):

src/main/java/competition/subsystems/coral_scorer/CoralScorerSubsystem.java
```diff
@@ -30,26 +35,41 @@ public CoralScorerSubsystem(XCANMotorController.XCANMotorControllerFactory xcanM
...
+            this.coralSensor = xDigitalInputFactory.create(electricalContract.getCoralSensor(),
+                    "CoralSensor");
+        } else {
+            this.coralSensor = null;
+        }
+
         this.intakePower = propertyFactory.createPersistentProperty("intakePower", .1);
         this.scorePower = propertyFactory.createPersistentProperty("scorerPower", -.1);
+
+        this.isCoralReady = propertyFactory.createPersistentProperty("isCoralReady", false);
```
aschokking, line 48: ⭐ ⭐ ⭐ this comes from the electricalContract, we don't want a property for it as well

aschokking (changes requested):

src/main/java/competition/subsystems/coral_scorer/CoralScorerSubsystem.java
```diff
@@ -30,26 +35,41 @@ public CoralScorerSubsystem(XCANMotorController.XCANMotorControllerFactory xcanM
...
+    public void scorer() {
+        setCoralScorerMotorPower(scorePower.get());
     }
-    public void stop(){
-        if (motor!=null){
-            motor.setPower(0);
+    public void stop() {
+        setCoralScorerMotorPower(0);
+    }
+    public boolean hasCoral() {
```
aschokking, line 68: ⭐ ⭐ in the function `public void periodic()`, let's publish this value so we can see it with a nice name at the subsystem level
```java
akitLogger.record("coralPresent", this.hasCoral())
```

aschokking (changes requested):

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -62,6 +62,13 @@ public CANMotorControllerInfo getArmPivotMotor() {
                 new CANMotorControllerOutputConfig());
     }
 
+    public boolean isCoralSensorReady() { return false; }
```
aschokking, line 65: ⭐ let's mark it ready in the UnitTestContract so we can use it in the simulator

aschokking (approved):

JohnGilb (approved):

PR #47, opened 2025-01-26
Arm maintainer

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

Created min & max values for arm maintainer.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/arm_pivot/commands/ArmPivotMaintainerCommand.java
+++ b/src/main/java/competition/subsystems/arm_pivot/commands/ArmPivotMaintainerCommand.java
@@ -7,7 +7,9 @@ import competition.subsystems.arm_pivot.ArmPivotSubsystem;
 import edu.wpi.first.units.measure.Angle;
 import xbot.common.command.BaseMaintainerCommand;
 import xbot.common.logic.HumanVsMachineDecider;
+import xbot.common.math.MathUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/competition/subsystems/arm_pivot/commands/ArmPivotMaintainerCommand.java
```diff
@@ -61,7 +67,8 @@ protected double getErrorMagnitude() { //distance from goal
 
     @Override
     protected double getHumanInput() { //gamepad controls: Left joy stick up/down & Left bumper to switch between elevator/arm
-       return oi.programmerGamepad.getLeftStickY();
+       return MathUtils.constrainDouble(oi.programmerGamepad.getLeftStickY(),
```
aschokking, line 70: ⭐  we should use the deadband untils too while we're here, see the Elevator's getHumanInput()

aschokking (approved):

src/main/java/competition/subsystems/arm_pivot/commands/ArmPivotMaintainerCommand.java
aschokking: right now you'll be only using a tiny range of the joystick before hitting your max

PR #48, opened 2025-01-26
Arm angle wrapping

kujo27:
# Why are we doing this?
Handles arm angle wrapping with arm pivot absolute encoder and low sensor
Asana task URL:

# Whats changing?
ArmPivotIsReady in ElectricalContract and Contract2025, ArmPivotSubsystem, new Test

# How this was tested
- [ 10 ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 

src/main/java/competition/electrical_contract/UnitTestContract2025.java
```diff
@@ -24,4 +24,5 @@ public UnitTestContract2025() {}
     public boolean areCanCodersReady() { return true; }
 
     public boolean isElevatorBottomSensorReady() { return true; }
+
```
aschokking, line 27: ⭐ ⭐ ⭐ please mark all devices as ready here in the unit test contract so we can exercise the code paths in the simulator

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -17,30 +19,47 @@
 @Singleton
 public class ArmPivotSubsystem extends BaseSetpointSubsystem<Angle> implements DataFrameRefreshable {
     public final XCANMotorController armMotor;
+    private final XAbsoluteEncoder armAbsoluteEncoder;
```
aschokking, line 22: ⭐ not a huge deal, but private is pretty harsh, I can't think of a lot of good reasons to make some more restricted than `protected`

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -17,30 +19,47 @@
...
+    private final XDigitalInput lowSensor;
     Angle targetAngle;
     ElectricalContract electricalContract;
     DoubleProperty degreesPerRotations;
     double rotationsAtZero;
     boolean isCalibrated = false;
+    double positionToDegrees = 125.0;
+    DoubleProperty minArmPosition;
```
aschokking, line 30: ⭐ these should all be final too

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -17,30 +19,47 @@
...
+                             XDigitalInput.XDigitalInputFactory xDigitalInputFactory) {
 
         propertyFactory.setPrefix(this);
         this.electricalContract = electricalContract;
-        if (electricalContract.isArmPivotMotorReady()) {
+
+        if (electricalContract.isArmPivotReady()) {
             this.armMotor = xcanMotorControllerFactory.create(electricalContract.getArmPivotMotor(),
                     getPrefix(), "ArmPivotMotor");
+            this.armAbsoluteEncoder = xAbsoluteEncoderFactory.create(electricalContract.getArmPivotAbsoluteEncoder(),
```
aschokking, line 45: ⭐ ⭐ ⭐ pull from `main`, there's a new method you want to call here that will make sure data frame refresh gets called on these devices

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -74,9 +93,53 @@ protected boolean areTwoTargetsEquivalent(Angle target1, Angle target2) {
         return target1.isEquivalent(target2);
     }
 
+    public Angle getArmAngle() {
+        double absoluteEncoderPosition = armAbsoluteEncoder.getAbsolutePosition().in(Degrees);
```
aschokking, line 97: ⭐ ⭐ ⭐ you need to guard here against the `armAbsoluteEncoder` possibly being null because the system isn't ready in the elec contract. Just return a 0 angle maybe in that case?

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -74,9 +93,53 @@ protected boolean areTwoTargetsEquivalent(Angle target1, Angle target2) {
...
+                armPosition = 1 - (maxArmPosition.get() - absoluteEncoderPosition);
+            }
+        }
+        // convert from [0,1] position to arm angle in degrees.
+        return Degrees.of(armPosition * positionToDegrees);
+    }
+
+    public static Angle getArmAngleTest(double minPosition, double maxPosition,
+                                        double absEncoderPosition, boolean sensorHit, double positionToDegrees) {
+        double armPosition = 0;
```
aschokking, line 121: ⭐ ⭐ ⭐ rather than duplicating all the logic here, have your other method call this one, providing all the values by reading the sensors.

The reason is that again someone could change the above logic in a way that introduces a bug that should cause the test tofail, but it won't because the test is only exercising this copy of the code

src/test/java/competition/subsystems/arm/ArmPivotAngleWrappingTest.java
```diff
@@ -0,0 +1,51 @@
...
+import competition.subsystems.arm_pivot.ArmPivotSubsystem;
+import org.junit.Test;
+
+import static edu.wpi.first.units.Units.Degrees;
+import static org.junit.Assert.assertEquals;
+
+public class ArmPivotAngleWrappingTest extends BaseCompetitionTest {
+    @Test
+    public void test() {
+        // Test #1: Absolute encoder position is in safe range, > maxPosition , < minPosition
```
aschokking, line 13: ⭐ instead of having these commends group tests, consider making separate test functions like
```java
@Test
public void testAbsoluteEncoderInSafeRange() { 
```

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -17,30 +19,47 @@
...
+            this.lowSensor = xDigitalInputFactory.create(electricalContract.getArmPivotLowSensor(),
+                    "ArmPivotLowSensor");
         } else {
             this.armMotor = null;
+            this.armAbsoluteEncoder = null;
+            this.lowSensor = null;
         }
 
         this.degreesPerRotations = propertyFactory.createPersistentProperty("Degrees Per Rotations", 1);
+        this.minArmPosition = propertyFactory.createPersistentProperty("Min AbsEncoder Position", 0.25);
```
aschokking, line 56: ⭐ consider saying what the units of these are in the property string, I'd have no idea at first blush

aschokking (changes requested):
Great start

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -74,4 +96,35 @@ public boolean isCalibrated() {
     protected boolean areTwoTargetsEquivalent(Angle target1, Angle target2) {
         return target1.isEquivalent(target2);
     }
+
+    public Angle getArmAngle() {
+        // if armAbsoluteEncoder is null, return 0
+        if (armAbsoluteEncoder.getAbsolutePosition() == null) {
```
aschokking, line 102: ⭐ ⭐shouldn't this be checking for the contract sensors being ready?

aschokking (approved):

src/test/java/competition/subsystems/arm/ArmPivotAngleWrappingTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertEquals(Degrees.of(31.25),
+                ArmPivotSubsystem.getArmAngle(0.75, 0.8,
+                        Degrees.of(360), false, 125));
+        assertEquals(Degrees.of( 100),
+                ArmPivotSubsystem.getArmAngle(0.75, 0.8,
+                        Degrees.of(216), false, 125));
+        assertEquals(Degrees.of( 62.5),
+                ArmPivotSubsystem.getArmAngle(0.75, 0.8,
+                        Degrees.of(108), false, 125));
+    }
```
JohnGilb, line 35: What's missing is a test where, for example, min is 0.99 and max is 0.04 (could happen depending on sensor mounting, so min is 0.99, and as you rotate the arm positively, increases from 0.99->1->0, goes through a full rotation to 1 and then 0 again, and a final bit to 0.04. That would represent 1.05 rotations, but it crosses the 0->1 barrier 2x.

JohnGilb (changes requested):

PR #53, opened 2025-01-29
Fix DebugSwerveModuleCommand

stephenjust:
# Whats changing?
Two primary fixes -
1. Require the drive subsystem
2. Apply a deadband on the joysticks

# How this was tested
- [ ] unit tests added
- [x] tested on robot

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/DebugSwerveModuleCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/DebugSwerveModuleCommand.java
@@ -26,9 +26,8 @@ public class DebugSwerveModuleCommand extends BaseCommand {
 
     @Override
     public void execute() {
-        // Use a generous deadband to prevent surprise movements
-        double drivePower = MathUtil.applyDeadband(oi.programmerGamepad.getLeftStickY(), 0.15);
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/drive/commands/DebugSwerveModuleCommand.java
```diff
@@ -24,8 +26,9 @@ public void initialize() {
 
     @Override
     public void execute() {
-        double drivePower = oi.programmerGamepad.getLeftStickY();
-        double turnPower = oi.programmerGamepad.getRightStickX();
+        // Use a generous deadband to prevent surprise movements
+        double drivePower = MathUtil.applyDeadband(oi.programmerGamepad.getLeftStickY(), 0.15);
```
aschokking, line 30: ⭐ we've got a baked in value for a reasonable global driving deadband that you could probably use here instead of the hardcoded value `driverDeadband`

aschokking (approved):

src/main/java/competition/subsystems/drive/commands/DebugSwerveModuleCommand.java
stephenjust: Fixed.

JohnGilb (approved):

PR #54, opened 2025-01-29
Negative power prevention

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

Created negative power prevention

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -58,6 +61,12 @@ private Angle getMotorPosition() {
         return Rotations.of(0);
     }
 
+    public void  preventNegative() {
```
JohnGilb, line 64: remove this method, since you're not using it below.

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -70,7 +79,10 @@ public void setTargetValue(Angle value) {
 
     @Override
     public void setPower(double power) {
-        if (electricalContract.isArmPivotMotorReady()) {
+        if (electricalContract.isArmPivotMotorReady() && coralScorerSubsystem.hasCoral()) {
```
JohnGilb, line 82: This doesn't seem like the right condition; the ArmPivot shouldn't allow negative power when it's too close to the bottom of its range - so you're going to need some methods to find out where in the range the arm is.

JohnGilb (changes requested):

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -22,6 +24,8 @@ public class ArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
     double rotationsAtZero;
     boolean isCalibrated = false;
 
+    ElevatorSubsystem elevatorSubsystem;
```
JohnGilb, line 27: not needed, leftover from earlier code

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -58,6 +62,10 @@ private Angle getMotorPosition() {
         return Rotations.of(0);
     }
 
+    private Angle getCalibratedPosition() {
```
JohnGilb, line 65: Sorry to mislead you about this one - the `getMotorPositionFromZeroOffset` method already does what we were looking for, in that it does the offset calculation internally. You should remove this method.

src/main/java/competition/subsystems/arm_pivot/ArmPivotSubsystem.java
```diff
@@ -70,6 +78,9 @@ public void setTargetValue(Angle value) {
 
     @Override
     public void setPower(double power) {
+        if (getCalibratedPosition().in(Rotations) < 0) {
```
JohnGilb, line 81: here, use `getMotorPositionFromZeroOffset`

JohnGilb (changes requested):

JohnGilb (approved):

PR #56, opened 2025-01-29
CalibrateDriveCommand

kujo27:
# Why are we doing this?
Allows us to run a command to find motor rotations to find meters per rotations.
Asana task URL:

# Whats changing?

CalibrateDriveCommand
# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 

src/main/java/competition/subsystems/drive/commands/CalibrateDriveCommand.java
```diff
@@ -0,0 +1,82 @@
...
+
+        aKitLog.record("frontLeftModuleStartPosition", frontLeftModuleStartPosition);
+        aKitLog.record("frontRightModuleStartPosition", frontRightModuleStartPosition);
+        aKitLog.record("rearLeftModuleStartPosition", rearLeftModuleStartPosition);
+        aKitLog.record("rearRightModuleStartPosition", rearRightModuleStartPosition);
+    }
+
+    @Override
+    public void execute() {
+        // wait for modules to be zeroed
```
JohnGilb, line 54: zeroing the modules may move the drive motors a bit - you might want to do the value recording after the 2 second settle.

src/main/java/competition/subsystems/drive/commands/CalibrateDriveCommand.java
```diff
@@ -0,0 +1,82 @@
...
+        if (timer.hasElapsed(2)) {
+            drive.setTargetSwerveStates(new SwerveModuleState(power, new Rotation2d(0)));
+            double frontLeftModuleDeltaPosition = drive.getFrontLeftSwerveModuleSubsystem().getDriveSubsystem()
```
JohnGilb, line 57: this much repeated code is begging for a helper method like 
`private getDeltaPosition(SwerveModule module, double startingPosition)`

JohnGilb (approved):

src/main/java/competition/subsystems/drive/commands/CalibrateDriveCommand.java
```diff
@@ -0,0 +1,90 @@
...
+    @Inject
+    public CalibrateDriveCommand(DriveSubsystem drive) {
+        this.drive = drive;
+        this.addRequirements(drive);
+    }
+
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+        this.timer.restart();
```
JohnGilb, line 33: This is ok for a debugging command, but you should be aware that you should always call `XTimer.getFPGATimestamp()` on any code that would run during a match. This allows for deterministic logging and replay.

JohnGilb (approved):

PR #61, opened 2025-01-31
Log camera poses

stephenjust:
# Why are we doing this?

Asana task URL: https://app.asana.com/0/1208731987242938/1209190708009168/f

# Whats changing?
Optionally logging the Pose3d for each camera on the robot.

# How this was tested
- [ ] unit tests added
- [x] tested on robot

-----

PR feedback legend

 

aschokking (approved):
Would be great to test it in the simulator to see how things render. We might need to add a new 'robot' custom model to show, but could do 'cone' or something in the short-run

stephenjust:
> Would be great to test it in the simulator to see how things render. We might need to add a new 'robot' custom model to show, but could do 'cone' or something in the short-run

Cone works visually, but you can sadly only attach the vision targets to a robot or ghost (and it snaps to the middle of the model, rather than the pose)

PR #62, opened 2025-01-31
Defined arm known positions

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
+++ b/src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
@@ -19,7 +19,8 @@ import static edu.wpi.first.units.Units.Rotations;
 public class CoralArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
 
     public enum ArmGoals {
-        Score
+        Score,
+        HumanLoad
     }
 
     public final DoubleProperty scoreAngle;
--- a/src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
+++ b/src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
@@ -35,10 +35,6 @@ public class CoralArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
     @Inject
     public CoralArmPivotSubsystem(XCANMotorController.XCANMotorControllerFactory xcanMotorControllerFactory,
                                   ElectricalContract electricalContract, PropertyFactory propertyFactory) {
-
-        this.scoreAngle = propertyFactory.createPersistentProperty("Scoring Angle", -125);
-        this.humanLoadAngle = propertyFactory.createPersistentProperty("Human Loading Angle", 0);
-
         propertyFactory.setPrefix(this);
 
         this.electricalContract = electricalContract;
@@ -51,6 +47,10 @@ public class CoralArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
         }
 
         this.degreesPerRotations = propertyFactory.createPersistentProperty("Degrees Per Rotations", 1);
+
+        this.scoreAngle = propertyFactory.createPersistentProperty("Scoring Angle", -125);
+        this.humanLoadAngle = propertyFactory.createPersistentProperty("Human Loading Angle", 0);
+
     }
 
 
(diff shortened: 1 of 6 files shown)


src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
```diff
@@ -10,11 +10,22 @@
...
 @Singleton
-public class ArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
+public class CoralArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
+
+    public enum ArmGoals {
+        Score,
+        HumanLoad
+    }
+
+    public final Angle scoreAngle;
```
stephenjust, line 26: ⭐⭐ Should these be properties, so that we can tune them more easily?

stephenjust (changes requested):

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
```diff
@@ -68,6 +82,14 @@ public void setTargetValue(Angle value) {
         targetAngle = value;
     }
 
+    public void setTargetAngle(ArmGoals value) {
+        if (Objects.requireNonNull(value) == ArmGoals.Score) {
```
stephenjust, line 86: ⭐⭐ Consider a switch-case statement instead. Use both enum values, and set the default to a safe state.

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
```diff
@@ -10,11 +10,21 @@
...
+import java.util.Objects;
+
 import static edu.wpi.first.units.Units.Degrees;
 import static edu.wpi.first.units.Units.Rotations;
 
 @Singleton
-public class ArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
+public class CoralArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
+
+    public enum ArmGoals {
```
stephenjust, line 21: ⭐⭐⭐ You should have all possible states in this list, even if they're not referenced in code. (If you use a switch-case statement, they will be referenced.)

stephenjust (changes requested):

stephenjust (approved):

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
aschokking: I think these are the only 2 states we support right now?

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
```diff
@@ -37,6 +48,10 @@ public ArmPivotSubsystem(XCANMotorController.XCANMotorControllerFactory xcanMoto
         }
 
         this.degreesPerRotations = propertyFactory.createPersistentProperty("Degrees Per Rotations", 1);
+
+        this.scoreAngle = propertyFactory.createPersistentProperty("Scoring Angle", -125);
```
aschokking, line 52: ⭐ ⭐  I think this should be positive 125 probably? Since the arm goes 'up' from 0. If there's an inversion we need somewhere else to make that work we can do that change there

aschokking, line 52: ⭐ consider including the units (degrees) in the name of the property so people populating it know what their doing

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
aschokking: Here's what it might look like as you probably haven't messed with switch statements a lot so far in your Java journey.
```java
switch(value) {
  case Score -> setTargetValue(Degrees.of(scoreAngle.get()));
  case HumanLoad -> setTargetValue(Degrees.of(humanLoadAngle.get()));
}
```

@stephenjust is there any reason to support a default when the cases are fully enumerated? The compiler will complain in the future if a new value is added but not handled here?

aschokking (approved):

aschokking:
I'm going to merge this so we can start using things but please come back and make a follow-up PR with the requested changes

PR #64, opened 2025-01-31
Change Gradle logging on builds to make it easier to find failures

stephenjust:
stacktrace is not really useful, since it spits out internal gradle stacks.

# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

-----

PR feedback legend

 

aschokking (approved):
Certainly room for improvement, hope this helps

PR #68, opened 2025-02-01
Trapezoid motion control for elevator

aschokking:
# Why are we doing this?

To enable smoother control of the elevator leverage https://docs.wpilib.org/en/stable/docs/software/advanced-controls/controllers/trapezoidal-profiles.html to limit the velocity and acceleration of system changes.

The goal here is to make this really easy to use so it's trivial to put it in our 3 positional systems (elevator, coral arm, pivot arm).

Asana task URL:

# Whats changing?

- Adds a new wrapper class for the wpilib trapezoid profile so that we can tune it with our Property system and plug it into our current pid logic easily
- This could be used with an on rio PID for position (as it is here) or the positional setpoint could just as easily be set on the motor controller itself

- Should this live in the SetpointSubsystem instead perhaps? I just started in the maintainer because that's where the pid code was as I was hacking it in.

## Future work

- Could probably make an assisted factory for this so the PF doesn't need to be given in manually, lazy there to start
- Add this to the coral and algae arm systems too when the dust settles

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

Tested in simulator

## Here's slow constraints causing very slow exaggerated motion as a demo:

https://github.com/user-attachments/assets/b91f8d94-b275-4c8d-a92b-319b03867123

## Changing the max accel or velocity changing the shape of the curve:
image

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/motion/TrapezoidProfileManager.java
+++ b/src/main/java/competition/motion/TrapezoidProfileManager.java
@@ -31,7 +31,7 @@ public class TrapezoidProfileManager {
         goalState = new TrapezoidProfile.State(initialPosition, 0);
     }
 
-    public void setTargetValue(double targetValue, double currentValue, double currentVelocity) {
+    public void setTargetPosition(double targetValue, double currentValue, double currentVelocity) {
         // if the profile's constraints properties have changed, recompute the profile
         // there's maybe a better place to do this but this should be fine since setTarget will be called
         // over and over again
@@ -50,7 +50,7 @@ public class TrapezoidProfileManager {
     }
 
     // currently only doing position, but in theory this goal has a velocity associated with it too we could use
-    public double getCurrentPositionSetpoint() {
+    public double getRecommendedPositionForTime() {
         var setpoint = profile.calculate(XTimer.getFPGATimestampTime().minus(profileStartTime).in(Seconds), initialState, goalState);
         return setpoint.position;
     }
--- a/src/main/java/competition/motion/TrapezoidProfileManager.java
+++ b/src/main/java/competition/motion/TrapezoidProfileManager.java
@@ -1,0 +1,58 @@
+package competition.motion;
+
+import static edu.wpi.first.units.Units.Seconds;
+
+import edu.wpi.first.math.trajectory.TrapezoidProfile;
+import edu.wpi.first.units.measure.Time;
+import xbot.common.controls.sensors.XTimer;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.PropertyFactory;
+
+public class TrapezoidProfileManager {
+    
+    
+    final DoubleProperty maxVelocity;
+    final DoubleProperty maxAcceleration;
+    
+    TrapezoidProfile profile;
+    TrapezoidProfile.Constraints constraints;
+    TrapezoidProfile.State initialState;
+    TrapezoidProfile.State goalState;
+    Time profileStartTime = Seconds.zero(); 
+
+    public TrapezoidProfileManager(String name, PropertyFactory pf, double defaultMaxVelocity, double defaultMaxAcceleration, double initialPosition) { 
+        pf.setPrefix(name);
+        maxVelocity = pf.createPersistentProperty("maxVelocity", defaultMaxVelocity);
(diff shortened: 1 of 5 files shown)


src/main/java/competition/motion/TrapezoidProfileManager.java
```diff
@@ -0,0 +1,58 @@
...
+        maxVelocity = pf.createPersistentProperty("maxVelocity", defaultMaxVelocity);
+        maxAcceleration = pf.createPersistentProperty("maxAcceleration", defaultMaxAcceleration);
+        constraints = new TrapezoidProfile.Constraints(maxVelocity.get(), maxAcceleration.get());
+        profile = new TrapezoidProfile(constraints);
+        // initialize states to current value
+        initialState = new TrapezoidProfile.State(initialPosition, 0);
+        goalState = new TrapezoidProfile.State(initialPosition, 0);
+    }
+
+    public void setTargetValue(double targetValue, double currentValue, double currentVelocity) {
```
JohnGilb, line 34: :star: should this be `setTargetPosition` ?

src/main/java/competition/motion/TrapezoidProfileManager.java
```diff
@@ -0,0 +1,58 @@
...
+        // if the profile's constraints properties have changed, recompute the profile
+        // there's maybe a better place to do this but this should be fine since setTarget will be called
+        // over and over again
+        if(constraints.maxVelocity != maxVelocity.get() || constraints.maxAcceleration != maxAcceleration.get()) {
+            constraints = new TrapezoidProfile.Constraints(maxVelocity.get(), maxAcceleration.get());
+            profile = new TrapezoidProfile(constraints);
+        }
+
+        // if the target has changed, recompute the goal and current states
+        if(goalState.position != targetValue) {
```
JohnGilb, line 44: :star: :star: might need a "is this actually different enough" to avoid issues with double comparison (which would in turn waste some time in the logic below, especially if we added some extra tracing)

src/main/java/competition/motion/TrapezoidProfileManager.java
```diff
@@ -0,0 +1,58 @@
...
+            initialState = new TrapezoidProfile.State(currentValue, currentVelocity);
+            goalState = new TrapezoidProfile.State(targetValue, 0);
+            profileStartTime = XTimer.getFPGATimestampTime();
+        }
+
+    }
+
+    // currently only doing position, but in theory this goal has a velocity associated with it too we could use
+    public double getCurrentPositionSetpoint() {
```
JohnGilb, line 53: :star: maybe `getRecommendedPositionForCurrentTime()`? I get that this might be fed as the setpoint to another object, but this stuff isn't strictly tied to PID, so perhaps something more generic?

src/main/java/competition/motion/TrapezoidProfileManager.java
```diff
@@ -0,0 +1,58 @@
...
+    final DoubleProperty maxVelocity;
+    final DoubleProperty maxAcceleration;
+    
+    TrapezoidProfile profile;
+    TrapezoidProfile.Constraints constraints;
+    TrapezoidProfile.State initialState;
+    TrapezoidProfile.State goalState;
+    Time profileStartTime = Seconds.zero(); 
+
+    public TrapezoidProfileManager(String name, PropertyFactory pf, double defaultMaxVelocity, double defaultMaxAcceleration, double initialPosition) {
```
JohnGilb, line 23: :star: it feels a little odd to seed this with just a `initialPosition` but not also an initial target, since this class needs both to work in tandem. Maybe just omit `initialPosition` in the constructor?

JohnGilb (approved):

src/main/java/competition/motion/TrapezoidProfileManager.java
aschokking: The challenge for me was what happens if `getCurrentPositionSetpoint` is called before `setTargetValue`?

src/main/java/competition/motion/TrapezoidProfileManager.java
aschokking: I was mirroring our naming of things in setpointSubsystem, but you're right in this case this has a strong position bias that we should make explicit here. That said the position could be angle or distance (or some other funky thing)

PR #70, opened 2025-02-01
Created switch-cased statement for arm angles

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
+++ b/src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
@@ -109,7 +109,6 @@ public class CoralArmPivotSubsystem extends BaseSetpointSubsystem<Angle> {
         switch (value) {
             case Score -> setTargetValue(Degrees.of(scoreAngle.get()));
             case HumanLoad -> setTargetValue(Degrees.of(humanLoadAngle.get()));
-            default -> setTargetValue(Degrees.of(humanLoadAngle.get()));
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
```diff
@@ -106,10 +106,10 @@ public void setTargetValue(Angle value) {
...
 
     public void setTargetAngle(ArmGoals value) {
-        if (Objects.requireNonNull(value) == ArmGoals.Score) {
-            setTargetValue(Degrees.of(scoreAngle.get()));
-        } else {
-            setTargetValue(Degrees.of(humanLoadAngle.get()));
+        switch (value) {
+            case Score -> setTargetValue(Degrees.of(scoreAngle.get()));
+            case HumanLoad -> setTargetValue(Degrees.of(humanLoadAngle.get()));
+            default -> setTargetValue(Degrees.of(humanLoadAngle.get()));
```
aschokking, line 112: ⭐  if we can omit the default case, that'll help protect us from future changes better I think (but I'm not 100% sure on the Java support for that)

aschokking (approved):

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
aschokking: Ah right I guess it could be null, nvm, this is fine

PR #73, opened 2025-02-01
Add SysId routines for the Swerve drive

stephenjust:
# Why are we doing this?
We need a practical SysId example to try it out.

Asana task URL:

# Whats changing?
Add a SysId gamepad to allow us to tune the pids for the robot drive.

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/operator_interface/OperatorInterface.java
+++ b/src/main/java/competition/operator_interface/OperatorInterface.java
@@ -17,6 +17,7 @@ import xbot.common.properties.PropertyFactory;
 public class OperatorInterface {
     public XXboxController gamepad;
     public XXboxController programmerGamepad;
+    public XXboxController sysIdGamepad;
 
     final DoubleProperty driverDeadband;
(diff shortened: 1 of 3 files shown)


src/main/java/competition/operator_interface/OperatorInterface.java
```diff
@@ -31,6 +32,10 @@ public OperatorInterface(XXboxControllerFactory controllerFactory, RobotAssertio
         programmerGamepad.setLeftInversion(false, true);
         programmerGamepad.setRightInversion(true, true);
 
+        sysIdGamepad = controllerFactory.create(5);
```
aschokking, line 35: ⭐ this name is going to conflict with David's, but we can cross that bridge when we come to it. If you want to update it to something in the middle now maybe call it `algaeAndSysIdGamepad`?

aschokking (approved):
Excited to try this!

PR #78, opened 2025-02-01
Created command to rotate arm to set angles

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/coral_arm_pivot/commands/CoralArmPivotMaintainerCommand.java
```diff
@@ -34,9 +35,10 @@ public CoralArmPivotMaintainerCommand(CoralArmPivotSubsystem armPivotSubsystem,
        this.oi = oi;
        pf.setPrefix(this);
        pf.setDefaultLevel(Property.PropertyLevel.Important);
+       positionalPID = pidf.create(getPrefix() + "PositionalPID", 0,0,0);
```
aschokking, line 38: ⭐ ⭐  default p should probably be something small but non-zero, otherwise if we lose our properties the robot won't move at all. Maybe 0.01?

src/main/java/competition/subsystems/coral_arm_pivot/commands/CoralArmPivotMaintainerCommand.java
```diff
@@ -46,12 +48,19 @@ public void initialize() {
...
     }
 
+
     @Override
     protected void calibratedMachineControlAction() { //manages and runs pid
+        double power = positionalPID.calculate(armPivotSubsystem.getTargetValue().in(Degrees),
+                armPivotSubsystem.getCurrentValue().in(Degrees));
 
+        armPivotSubsystem.setPower(power);
+        aKitLog.record("Target Angle", armPivotSubsystem.getTargetValue().in(Degrees));
```
aschokking, line 61: ⭐ ⭐ this is duplicated with what's in the subsystem, can remove these 2?

aschokking (approved):
Looks great!

src/main/java/competition/subsystems/coral_arm_pivot/CoralArmPivotSubsystem.java
```diff
@@ -34,8 +34,8 @@ public enum ArmGoals {
     Angle targetAngle = Degrees.of(0);
     ElectricalContract electricalContract;
     DoubleProperty degreesPerRotations;
-    double rotationsAtZero;
-    boolean isCalibrated = false;
+    double rotationsAtZero = 0;
+    boolean isCalibrated = true;
```
JohnGilb, line 38: this is dangerous to commit. Instead, create a command that can force the arm to the calibrated state (check the elevator subsystem, it has a force calibration command)

JohnGilb (changes requested):

PR #85, opened 2025-02-01
Algae arm lower sensor

Glowdisk:
# Why are we doing this?
To detect it going to bottom
Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/algae_arm/AlgaeArmSubsystem.java
```diff
@@ -36,6 +41,12 @@ public AlgaeArmSubsystem(ElectricalContract electricalContract,
         } else{
             this.armMotor=null;
         }
+
+        if (electricalContract.isAlgaeArmBottomSensorReady()){
+            this.bottomSensor= xDigitalInputFactory.create(electricalContract.getElevatorBottomSensor(), "ElevatorBottomSensor");
```
JohnGilb, line 46: instead of `"ElevatorBottomSensor"`, this needs to be `this.getPrefix()`

JohnGilb (changes requested):

JohnGilb (approved):

src/main/java/competition/subsystems/algae_arm/AlgaeArmSubsystem.java
```diff
@@ -72,6 +83,13 @@ public void setPower(double power) {
 
     }
 
+    public boolean isTouchingBottom(){
```
aschokking, line 86: ⭐ ⭐ consider logging this to aKitLog.record in this classes `periodic()` call like is happening in the elevator for that bottom sensor (you might have to add the method in the first place)

aschokking (approved):

PR #92, opened 2025-02-02
Add a test that will verify that all subsystems with motor controller…

stephenjust:
…s call periodic on those motor controllers.

# Why are we doing this?
If we don't call periodic on motor controllers in a subsystem, we won't get properties emitted in NetworkTables.

# Whats changing?
Add a test that gets all of the injected subsystems by reflection and checks if motor controllers have periodic called.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [x] unit tests added

-----

PR feedback legend

 

aschokking (approved):
Super cool

PR #94, opened 2025-02-02
 added constant power to counteract gravity in Sim, + Deleting unneeded stuff in maintainer

junyu101:
# Why are we doing this?
To apply minimum power in the sim so elevator doensn't need to constantly pid towards it because of gravity
Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
+++ b/src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
@@ -5,10 +5,8 @@ import competition.operator_interface.OperatorInterface;
 import competition.subsystems.elevator.ElevatorSubsystem;
 import edu.wpi.first.units.measure.Distance;
 import xbot.common.command.BaseMaintainerCommand;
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -94,7 +86,8 @@ protected void calibratedMachineControlAction() {
         double power = positionPID.calculate(
                 setpoint,
                 elevator.getCurrentValue().in(Meters));
-        elevator.setPower(power);
+        //we dont need to counteract gravity when moving down
```
aschokking, line 89: A reasonable idea! This does create a little bit of a power cliff around 0 that we might want to smooth but we can look into that later on the real robot if it's a problem

aschokking (approved):
Great cleanup!

PR #98, opened 2025-02-04
Enums to pose

anthonytrenh:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

JohnGilb (approved):
Very nice!

Just as an FYI, take a look at https://github.com/Team488/TeamXbot2025/blob/main/src/main/java/competition/subsystems/oracle/ReefCoordinateGenerator.java, we may be able to generate a lot of the "Reef" positions procedurally instead of managing a handcrafted list. That's a longer-term goal, however, and the approach we have here has some advantages as well (easier to tune Landmarks)

aschokking (approved):

PR #99, opened 2025-02-04
Oracle: sub state machine class proof of concept

aschokking:
This is a hacky proof of concept of trying to encapsulate sub state machines of the oracle. It's not complete/meant to work now, the constructor isn't implemented at all for instance.

The idea is that these sub state machines don't know much about what the parent state machine is doing (the 'primary' state in the current implementation). That way the parent can change things around and hopefully the child can stay untouched. Conversely if the child state machine needs to be improved upon/made more complicated it can do so without the parent code having to change at all.

I also really wanted to have as few boolean flags as possible and instead use state enums to track when things are being initialized. 

I'd also envision splitting up more of what's currently in the OracleSubsystem so that the Primary state machine is 1 thing and then the holder of the all the current state/pointers to things is maybe another class that the sub state machines have access to when they need to set things?

### DIFF
--- a/src/main/java/competition/subsystems/oracle/CollectionStateMachine.java
+++ b/src/main/java/competition/subsystems/oracle/CollectionStateMachine.java
@@ -1,0 +1,68 @@
+package competition.subsystems.oracle;
+
+import java.util.List;
+
+import competition.subsystems.coral_scorer.CoralScorerSubsystem;
+import competition.subsystems.oracle.contracts.CoralCollectionInfoSource;
+import competition.subsystems.pose.Landmarks;
+import competition.subsystems.pose.PoseSubsystem;
+import edu.wpi.first.math.geometry.Pose2d;
+import xbot.common.advantage.AKitLogger;
+import xbot.common.trajectory.XbotSwervePoint;
+
+public class CollectionStateMachine {
+    public enum State {
+        INITIALIZING,
+        WAITING_FOR_CORAL,
+    }
+
+    public enum Result {
+        RUNNING,
+        CORAL_ACQUIRED,
+        FAILURE,
+    }
+
+    PoseSubsystem pose;
+    AKitLogger aKitLog;
+    OracleSubsystem oracle;
+    CoralCollectionInfoSource coralInfoSource;
+    
+    State state = State.INITIALIZING;
+    // Pose2d goalPose;
+
+    public void reset() {
+        state = State.INITIALIZING;
+    }
+
+    public Result run() {
+        switch (state) {
+            case INITIALIZING:
+                // TODO: see if we already have Coral, if so just end the machine now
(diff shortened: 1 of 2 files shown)


src/main/java/competition/subsystems/oracle/CollectionStateMachine.java
```diff
@@ -0,0 +1,68 @@
...
+                // goalPose = newDriveAdvice.path().get(newDriveAdvice.path().size() - 1).keyPose;
+                oracle.setCurrentDriveAdvice(newDriveAdvice);
+                // Command the superstructure
+                var newSuperstructureAdvice = new OracleSuperstructureAdvice(
+                        oracle.getNextInstructionNumber(), Landmarks.CoralLevel.COLLECTING, CoralScorerSubsystem.CoralScorerState.INTAKING);
+                oracle.setSuperstructureAdvice(newSuperstructureAdvice);
+                state = State.WAITING_FOR_CORAL;
+                break;
+            case WAITING_FOR_CORAL:
+                if(coralInfoSource.confidentlyHasCoral()) {
```
aschokking, line 53: was thinking this maybe even is outside the state machine loop because if it's ever true the machine can just end?

src/main/java/competition/subsystems/oracle/OracleSubsystem.java
```diff
@@ -214,6 +208,11 @@ private void setNextPrimaryActivity(PrimaryActivity activity) {
         firstRunInPrimaryActivity = true;
     }
 
+    private void setPrimaryActivityCollection() {
```
JohnGilb, line 211: What calls this?

JohnGilb:
This is a nice organization - gets the minimum set of stuff done without a ton of overhead. If the state machine implemented a "Resettable" interface, then when you're writing code in the Oracle you wouldn't have to specifically remember to call reset on the next state's object, as that could be done via a lookup of all registered state machines.

src/main/java/competition/subsystems/oracle/OracleSubsystem.java
aschokking: Ah this should be called from the primary state machine loop when it switches to Collection state

aschokking: There's some nuance about how to run reset() only when the state switches vs constantly in that state that I need to think about more.

aschokking: Maybe a latch?

aschokking: Just updated the code, I remember now I wanted to call it everywhere setting the activity to collection was happening. I think a latch or something might be cleaner and allow setting it again as a no-op

PR #103, opened 2025-02-05
IntakeUntill coral command made!

Glowdisk:
Using previous knowledge!

# Why are we doing this?

to make it run while being anonymous

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/coral_scorer/commands/IntakeUntillCoral.java
```diff
@@ -0,0 +1,28 @@
...
+import competition.subsystems.coral_scorer.CoralScorerSubsystem;
+import xbot.common.command.BaseCommand;
+
+import javax.inject.Inject;
+
+public class IntakeUntillCoral extends BaseCommand {
+    CoralScorerSubsystem coral;
+
+    @Inject
+
```
aschokking, line 12: ⭐ nit remove this extra empty line so the inject is beside the constructor.

src/main/java/competition/subsystems/coral_scorer/commands/IntakeUntillCoral.java
```diff
@@ -0,0 +1,28 @@
...
+    public IntakeUntillCoral (CoralScorerSubsystem coralScorerSubsystem){
+        coral= coralScorerSubsystem;
+        this.addRequirements(coral);
+    }
+
+    @Override
+    public void initialize() {
+        if (this.coral.confidentlyHasCoral()) {
```
aschokking, line 20: ⭐ ⭐ ⭐ you're missing an `isFinished()` method so that this command can end when the coral is present.

You could zero out the power there instead of here when the command is finished too

aschokking (changes requested):

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
```diff
@@ -0,0 +1,44 @@
...
+import edu.wpi.first.wpilibj2.command.ParallelCommandGroup;
+import edu.wpi.first.wpilibj2.command.ParallelDeadlineGroup;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import xbot.common.command.BaseCommand;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+public class HumanLoadUntillCoralCollectedCommandGroup extends ParallelCommandGroup {
+    Provider<SetCoralArmTargetAngleCommand> setCoralArmTargetAngleCommandProvider;
```
aschokking, line 17: ⭐ ⭐ these 3 variables are unused (and are just `null` because nothings ever assigned to them), you can safely remove them

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
```diff
@@ -0,0 +1,44 @@
...
+    Provider<SetElevatorTargetHeightCommand> setElevatorTargetHeightCommandProvider;
+    Provider<IntakeUntillCoral> intakeUntillCoralProvider;
+
+    @Inject
+    public HumanLoadUntillCoralCollectedCommandGroup(Provider<SetCoralArmTargetAngleCommand> setCoralArmTargetAngleCommandProvider,
```
aschokking, line 22: ⭐ Since you never need multiple copies of any of these commands, you could safely drop the Provider part and just ask for the Command directly

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
```diff
@@ -0,0 +1,44 @@
...
+                                                     Provider<SetElevatorTargetHeightCommand> setElevatorTargetHeightCommandProvider,
+                                                     Provider<IntakeUntillCoral> intakeUntillCoralProvider){
+        var setArmtoCollect= setCoralArmTargetAngleCommandProvider.get();
+        setArmtoCollect.setAngle(Landmarks.CoralLevel.COLLECTING);
+        var setElevatorTargetHeight= setElevatorTargetHeightCommandProvider.get();
+        setElevatorTargetHeight.setHeight(Landmarks.CoralLevel.COLLECTING);
```
aschokking, line 28: ⭐ ⭐ there's already a CommandGroup that Anthony made to do these 2 things together you should leverage, checkout `PrepCoralSystemCommandGroupFactory.create()`

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
```diff
@@ -0,0 +1,44 @@
...
+        var setIntakeUntillCoral= intakeUntillCoralProvider.get();
+        setIntakeUntillCoral.isFinished();
```
aschokking, line 30: ⭐ ⭐ ⭐ you don't need to call isFinished() yourself, and in fact this is kind of dangerous because it could be setting some state you didn't intend to. All the command state machine methods like `initialize()` `execute()` and `isFinished()` you never have to call yourself directly. The Scheduler calls these for you on running commands according to this state machine logic: https://github.com/Team488/SeriouslyCommonLib/wiki/Commands

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
```diff
@@ -0,0 +1,44 @@
...
+
+        this.addCommands((Command) intakeUntillCoralProvider,
+                (Command) setCoralArmTargetAngleCommandProvider,
```
aschokking, line 33: ⭐ ⭐ ⭐ because these commands never finish, your whole ParallelCommandGroup will never finish. On Saturday let's talk about the concept of "deadlines" as it related to ParallelCommandGroups

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
```diff
@@ -0,0 +1,44 @@
...
+                (Command) setElevatorTargetHeightCommandProvider);
+
+        //this.setCoralArmTargetAngleCommandProvider=setCoralArmTargetAngleCommandProvider;
```
aschokking, line 36: ⭐ please delete unused code like this

src/main/java/competition/subsystems/coral_scorer/commands/IntakeUntillCoral.java
```diff
@@ -0,0 +1,28 @@
...
+    public void initialize(){}
+
+    @Override
+    public boolean isFinished() {
+        if (this.coral.confidentlyHasCoral()) {
+            coral.motor.setPower(0);
+        } else {
+            coral.motor.setPower(coral.intakePower.get());
+        }
+        return true;
```
aschokking, line 26: ⭐ ⭐ ⭐ by always returning `true` here, your command is going to instantly end even if the robot doesn't have any coral present yet. Can you update this method so that it returns true if we have coral and false if we don't?

aschokking (changes requested):
Good progress, there's a lot of new concepts here related to command groups and things so I'm glad you've decided to tackle this task. Let's talk through some of these concepts more in person on Saturday!

src/main/java/competition/commandgroups/HumanLoadUntillCoralCollectedCommandGroup.java
anthonytrenh: changed setCoralArm and setElevator to prepCoralSystem which ends

src/main/java/competition/subsystems/coral_scorer/commands/IntakeUntillCoral.java
```diff
@@ -0,0 +1,29 @@
...
+public class IntakeUntillCoral extends BaseCommand {
+    CoralScorerSubsystem coral;
+
+    @Inject
+    public IntakeUntillCoral (CoralScorerSubsystem coralScorerSubsystem){
+        coral= coralScorerSubsystem;
+        this.addRequirements(coral);
+    }
+    @Override
+    public void initialize(){}
```
aschokking, line 17: ⭐ I'd suggest setting the intake power here in init(), but I think doing it in isFinished as you are is functionally okay

aschokking (approved):

PR #104, opened 2025-02-05
Simulating CoralArmPivotSubsystem Absolute Encoder

kujo27:
# Why are we doing this?
Allows us to simulate absolute encoder logic
Asana task URL:
https://app.asana.com/0/1208731987242938/1209287440592554
# Whats changing?
CoralArmPivotSubsystem, CoralArmSimulator, new CoralArmAngleSimulatorTest
Need PR for setPosition_interval in SCL for code to build.
# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/simulation/coral_arm/CoralArmSimulator.java
```diff
@@ -59,6 +65,14 @@ public void update() {
 
         var armMotorRotations = armRelativeAngle.in(Radians) / CoralArmSimConstants.armEncoderAnglePerRotation.in(Radians);
         armMotor.setPosition(Rotations.of(armMotorRotations));
+
+        absoluteEncoder.setPosition_internal(getAbsoluteEncoderPosition(getArmAngle(), armPivotSubsystem.minArmPosition.get() / 360
```
aschokking, line 69: ⭐ ⭐ ⭐ I think your PR is missing updating SCL to get your change to add this method right?

src/main/java/competition/simulation/coral_arm/CoralArmSimulator.java
```diff
@@ -70,6 +84,31 @@ public Angle getArmAngle() {
     }
 
     public boolean isAtCollectionAngle() {
-        return getArmAngle().isNear(Degrees.of(0), Degrees.of(4));
+        return getArmAngle().isNear(Degrees.of(0), 0.05);
```
aschokking, line 87: ⭐ ⭐ ⭐ it's actually really important that the second argument to `isNear` be a unit-ed value and not just a double, but especially when the first value is 0. That's because when you provide just a double it says the error margin is the first value times the double as a percentage. So `0*0.05 == 0` which is not a useful margin here.

src/main/java/competition/simulation/coral_arm/CoralArmSimulator.java
```diff
@@ -70,6 +84,31 @@ public Angle getArmAngle() {
...
     }
-}
+
+    public static Angle getAbsoluteEncoderPosition(Angle armAngle, double minPosition, double maxPosition) {
```
aschokking, line 90: ⭐ at first blush, I'm not sure what max and min position are, could you add a comment about them or have a more descriptive name?

aschokking: Like could this be unit-ed instead of doubles? Is it rotation counts? Degrees?

src/test/java/competition/subsystems/arm/CoralArmAngleSimulatorTest.java
```diff
@@ -0,0 +1,87 @@
...
+import competition.BaseCompetitionTest;
+import competition.simulation.coral_arm.CoralArmSimulator;
+
+import competition.subsystems.coral_arm_pivot.CoralArmPivotSubsystem;
+import org.junit.Test;
+
+import static edu.wpi.first.units.Units.Degrees;
+import static org.junit.Assert.assertEquals;
+
+public class CoralArmAngleSimulatorTest extends BaseCompetitionTest {
```
aschokking, line 12: 👏🏻 yay tests!

aschokking (changes requested):

src/main/java/competition/simulation/coral_arm/CoralArmSimulator.java
```diff
@@ -84,9 +82,11 @@ public Angle getArmAngle() {
     }
 
     public boolean isAtCollectionAngle() {
-        return getArmAngle().isNear(Degrees.of(0), 0.05);
+        return getArmAngle().isNear(Degrees.of(0), Degrees.of(4));
     }
 
+    // minPosition and maxPosition are the absolute encoder minimum and maximum position in degrees
```
aschokking, line 88: ⭐  this isn't degrees then right, rather rotations?

aschokking (approved):

PR #105, opened 2025-02-05
PID Via Motor Controllers & Power limits near extreme heights (This is Junyu)

anthonytrenh:
# Why are we doing this?
For smoother motion in elevator 

Asana task URL:

# Whats changing?
Just small changes to elevator maintainer modifying how power is set

# How this was tested
- [ ] tested on robot
- [x ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
+++ b/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
@@ -50,6 +50,8 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
     public Distance elevatorTargetHeight;
 
     public final DoubleProperty metersPerRotation;
+    public final DoubleProperty rotationsPerMeter;
+
     public final DoubleProperty calibrationNegativePower;
     public final DoubleProperty upperHeightLimit;
     public final DoubleProperty lowerHeightLimit;
@@ -94,8 +96,12 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
 
 
         //to be tuned
-        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/1923.0);
+        this.rotationsPerMeter = pf.createPersistentProperty("RotationsPerMeter", 1923.0);
+        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/rotationsPerMeter.get());
+
         this.calibrationNegativePower = pf.createPersistentProperty("calibrationNegativePower", -0.05);
+
+        //power limits near max and min height
         this.upperHeightLimit = pf.createPersistentProperty("upperHeightLimit", l4Height.get());
         this.lowerHeightLimit = pf.createPersistentProperty("lowerHeightLimit", baseHeight.get());
         this.powerNearUpperLimitThreshold = pf.createPersistentProperty("powerNearUpperLimit", 0.0);
--- a/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
+++ b/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
@@ -51,8 +51,8 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
 
     public final DoubleProperty metersPerRotation;
     public final DoubleProperty calibrationNegativePower;
-    public final DoubleProperty nearUpperLimitThreshold;
-    public final DoubleProperty nearLowerLimitThreshold;
+    public final DoubleProperty upperHeightLimit;
+    public final DoubleProperty lowerHeightLimit;
     public final DoubleProperty powerNearLowerLimitThreshold;
     public final DoubleProperty powerNearUpperLimitThreshold;
     public final DoubleProperty powerWhenBottomSensorHit;
@@ -84,14 +84,6 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
         this.elevatorTargetHeight = Inches.of(0);
 
         pf.setPrefix(this);
-        //to be tuned
-        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/1923.0);
-        this.calibrationNegativePower = pf.createPersistentProperty("calibrationNegativePower", -0.05);
-        this.nearUpperLimitThreshold = pf.createPersistentProperty("nearUpperLimitThreshold", 1.0);
-        this.nearLowerLimitThreshold = pf.createPersistentProperty("nearLowerLimitThreshold", 0.25);
-        this.powerNearUpperLimitThreshold = pf.createPersistentProperty("powerNearUpperLimit", 0.1);
-        this.powerNearLowerLimitThreshold = pf.createPersistentProperty("powerNearLowerLimit", -0.1);
-        this.powerWhenBottomSensorHit = pf.createPersistentProperty("powerWhenBottomSensorHit", -0.01);
 
         //these are not real measured heights yet, just placeholders
         l2Height = pf.createPersistentProperty("l2Height-m", Inches.of(1).in(Meters));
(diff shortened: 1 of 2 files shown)


src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -100,6 +94,20 @@ public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFac
...
 
+
+        //to be tuned
+        this.rotationsPerMeter = pf.createPersistentProperty("RotationsPerMeter", 1923.0);
+        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/rotationsPerMeter.get());
+
+        this.calibrationNegativePower = pf.createPersistentProperty("calibrationNegativePower", -0.05);
+
+        //power limits near max and min height
+        this.upperHeightLimit = pf.createPersistentProperty("upperHeightLimit", l4Height.get());
```
aschokking, line 105: ⭐ ⭐ please state the units in the variable + property string name. Or alternatively we just added a `DistanceProperty` that you can use for this

aschokking, line 105: ⭐ we should ask #cadawans if the elevator can maybe go a little bit above this point, otherwise we'll have no overhead for PIDing there with a tiny overshoot

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -114,6 +122,7 @@ public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFac
 
         if(contract.isElevatorReady()){
             this.masterMotor = motorFactory.create(contract.getElevatorMotor(), this.getPrefix(), "ElevatorMotor");
+            masterMotor.setPidDirectly(1,0,0.5,0,0.1);
```
aschokking, line 125: ⭐ ⭐ ⭐ instead of doing this, you'll want to pass the PID properties into the create() call above, there's an optional extra arg for setting the defaults (and from there they can be tweaked via the property system just like with PIDManager)

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -199,11 +208,11 @@ public boolean isTouchingBottom(){
     }
 
     public boolean isNearUpperLimit(){
-        return getCurrentValue().in(Meters) > nearUpperLimitThreshold.get();
+        return getCurrentValue().in(Meters) > upperHeightLimit.get();
```
aschokking, line 211: ⭐ ⭐ this is called `isNearUpperLimit` but this logic is saying we're past the upper limit. Could you see how someone might find that disconnect confusing? Maybe `upperHeightLimit` shouldn't be called the limit if this isn't the actual upper limit but just the value 'near' it? Or conversely maybe this method should be renamed to say at/past upper limit?

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -85,11 +89,10 @@ protected void calibratedMachineControlAction() {
...
         aKitLog.record("elevatorProfileTarget", setpoint);
 
-        double power = positionPID.calculate(
-                setpoint,
-                elevator.getCurrentValue().in(Meters));
-        //we dont need to counteract gravity when moving down
-        elevator.setPower(power < 0 ? power : power + gravityPIDConstantPower.get());
+        //handles pidding via motor controller and setting power to elevator
+        elevator.masterMotor.setPositionTarget(
+                Degrees.of(setpoint * elevator.rotationsPerMeter.get() * 360),
```
aschokking, line 94: ⭐ ⭐ The unit you want to use here is `Rotations.of()`

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -85,11 +89,10 @@ protected void calibratedMachineControlAction() {
...
+                XCANMotorController.MotorPidMode.TrapezoidalVoltage);
```
aschokking, line 95: ⭐ I think to start this could just be `Voltage` because we're doing the trapezoidal profiling part already on the rio side

aschokking (changes requested):

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
junyu101: will do tonight!

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -84,21 +89,27 @@ public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFac
         this.elevatorTargetHeight = Inches.of(0);
 
         pf.setPrefix(this);
+
+        //these are not real measured heights yet, just placeholders
+        l2Height = pf.createPersistentProperty("l2Height-inches", Inches.of(1));
```
aschokking, line 94: ⭐ ⭐ turns out the DistanceProperty actually already attached the unit to the name for you because this is such a common problem! You can omit the "-inches" here and it will add it based on the unit type of the default value you provide

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -84,21 +89,27 @@ public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFac
...
+        l3Height = pf.createPersistentProperty("l3Height-inches", Inches.of(15.875));
+        l4Height = pf.createPersistentProperty("l4Height-inches", Inches.of(40.651));
+        humanLoadHeight = pf.createPersistentProperty("humanLoadHeight-inches", Inches.of(1));
+        baseHeight = pf.createPersistentProperty("baseHeight-inches", Inches.of(0));
+
+
         //to be tuned
-        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/1923.0);
+        this.rotationsPerMeter = pf.createPersistentProperty("RotationsPerMeter", 1923.0);
+        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/rotationsPerMeter.get());
```
aschokking, line 103: ⭐ ⭐ ⭐ we probably don't need properties for both of these values since 1 is derivable from the other via this inversion right?

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -113,7 +124,10 @@ public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFac
         );
 
         if(contract.isElevatorReady()){
-            this.masterMotor = motorFactory.create(contract.getElevatorMotor(), this.getPrefix(), "ElevatorMotor");
+            this.masterMotor = motorFactory.create(
+                    contract.getElevatorMotor(), this.getPrefix(), "ElevatorMotorPID",
+                    new XCANMotorControllerPIDProperties(1,0,0.5)
```
aschokking, line 129: 👏🏻

aschokking (changes requested):
Really close! Nice job on the edits so far. Last round probably 🤞🏻

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -51,8 +51,8 @@ public enum ElevatorPowerRestrictionReason{
 
     public Distance elevatorTargetHeight;
 
-    public final DoubleProperty metersPerRotation;
     public final DoubleProperty rotationsPerMeter;
+    public final double metersPerRotation;
```
aschokking, line 55: ⭐ this could still be a Distance

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -91,22 +91,22 @@ public ElevatorSubsystem(XCANMotorController.XCANMotorControllerFactory motorFac
...
+        l3Height = pf.createPersistentProperty("l3Height", Inches.of(15.875));
+        l4Height = pf.createPersistentProperty("l4Height", Inches.of(40.651));
+        humanLoadHeight = pf.createPersistentProperty("humanLoadHeight", Inches.of(1));
+        baseHeight = pf.createPersistentProperty("baseHeight", Inches.of(0));
 
 
         //to be tuned
         this.rotationsPerMeter = pf.createPersistentProperty("RotationsPerMeter", 1923.0);
-        this.metersPerRotation = pf.createPersistentProperty("MetersPerRotation", 1.0/rotationsPerMeter.get());
+        this.metersPerRotation = 1.0/ rotationsPerMeter.get();
```
aschokking, line 103: ⭐ ⭐ ⭐ I should have mentioned this before, but we need to guard against divide by 0 here or the robot could totally fail to do anything in a match 😨

aschokking (approved):

PR #108, opened 2025-02-07
Add to oracle scoring queue

Rongrrz:
# Why are we doing this?
So that the oracle scoring queue can be dynamic and modified in real time

Asana task URL:

# Whats changing?
A new command

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -53,14 +54,18 @@ public void setupDriverCommands(
...
         resetHeading.setHeadingToApply(0);
         operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.A).onTrue(resetHeading);
         operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.Y).whileTrue(alignToReefWithAprilTag);
 
         var oracleControlsRobot = Commands.parallel(driveAccordingToOracle, superstructureAccordingToOracle);
 
         operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.RightBumper).whileTrue(oracleControlsRobot);
+
+        addToOracleScoringQueue.setTaskToAdd(new ScoringTask(Landmarks.ReefFace.CLOSE, Landmarks.Branch.A, Landmarks.CoralLevel.FOUR));
+        operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.LeftBumper).onTrue(addToOracleScoringQueue);
```
aschokking, line 68: ⭐  you're going to have a conflict here in `main`, we'll need to figure out the right home for these, but probably the neotrellis 'gamepad' instead?

src/main/java/competition/subsystems/oracle/ScoringQueue.java
```diff
@@ -11,6 +11,7 @@
 public class ScoringQueue {
 
     private final Queue<ScoringTask> scoringTasks;
+    private boolean defaultState = true;
```
aschokking, line 14: ⭐ I think this is a reasonable short-run solution, but medium term I'm not sure if we'll have a 'default' state and instead it should probably always start empty and get populated as appropriate by the autonomous command setup stuff?

src/main/java/competition/subsystems/oracle/commands/AddToOracleScoringQueueCommand.java
```diff
@@ -0,0 +1,36 @@
...
+import competition.subsystems.oracle.ScoringTask;
+import xbot.common.command.BaseCommand;
+import xbot.common.logging.RobotAssertionManager;
+
+import javax.inject.Inject;
+
+public class AddToOracleScoringQueueCommand extends BaseCommand {
+
+    RobotAssertionManager assertionManager;
+    OracleSubsystem oracle;
```
aschokking, line 13: ⭐ nit, these should both be `final`

aschokking (approved):

src/main/java/competition/operator_interface/OperatorCommandMap.java
Rongrrz: I'll remove this for now until the OI gets figured out :)

src/main/java/competition/subsystems/oracle/ScoringQueue.java
Rongrrz: I'll keep it here for now for the case where we will have a default, if we don't have a default, this wouldn't hurt and can always be removed later.

JohnGilb (approved):

PR #112, opened 2025-02-08
Simulate lights

aschokking:
# Why are we doing this?

To help test + debug the lightSubsystem, it would be nice if we could visualize the different states when simulating the robot. This also has the added benefit of helping simulated human drivers start to learn how to read the lights.

Asana task URL:

# Whats changing?

- New LightsSimulator

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

Here's the robot starting disconnected from DS and then I toggle enabled/disabled back and forth:

https://github.com/user-attachments/assets/ab49809b-d1a7-451a-b302-08b57ae1b67e

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/simulation/elevator/ElevatorSimulator.java
+++ b/src/main/java/competition/simulation/elevator/ElevatorSimulator.java
@@ -23,6 +23,7 @@ import xbot.common.controls.actuators.mock_adapters.MockCANMotorController;
 import xbot.common.math.PIDManager;
 import xbot.common.math.PIDManager.PIDManagerFactory;
 import xbot.common.properties.PropertyFactory;
+import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj.MockDigitalInput;
 
 @Singleton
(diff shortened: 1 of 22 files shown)


src/main/java/competition/simulation/elevator/ElevatorSimulator.java
```diff
@@ -69,7 +70,11 @@ public Distance getCurrentHeight() {
     public void update() {
         MotorInternalPIDHelper.updateInternalPID(motor, pidManager);
 
-        this.elevatorSim.setInputVoltage(this.motor.getPower() * RobotController.getBatteryVoltage());
+        if(DriverStation.isEnabled()) {
```
aschokking, line 73: we actually need to go and do this to all the simulated mechanisms since currently it's still trying to set power while disabled. Or we need to implement that logic inside the MockCANMotor controllers but that seems like a lot more complexity

stephenjust (approved):

JohnGilb (approved):

PR #121, opened 2025-02-08
Created motor controller PID for coral arm

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

Note: human input is not working but is work in progress

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/coral_arm/CoralArmSubsystem.java
```diff
@@ -83,16 +87,20 @@ public CoralArmPivotSubsystem(XCANMotorController.XCANMotorControllerFactory xca
 
     @Override
     public Angle getCurrentValue() {
-        double currentAngle = calibratedPosition().in(Rotations) * degreesPerRotations.get();
-        return Degrees.of(currentAngle);
+        Angle currentAngle = Degrees.of(0);
+        if (electricalContract.isCoralArmMotorReady()) {
+            currentAngle = Degrees.of(
+                            (calibratedPosition().in(Rotations)) * degreesPerRotations.in(Degrees));
```
stephenjust, line 93: Does this work? Angle has a `.times()` method on it that might be useful here?

stephenjust: I'm thinking you might be able to keep the similar math in the Angles entirely without converting to doubles to do thae math.

AlanYuan99: Works normally, I think the .times() method makes it not work.

stephenjust (approved):

PR #123, opened 2025-02-08
Add missing guards in elevator code

stephenjust:
# Why are we doing this?
We need to guard against missing elevator components so that robots don't crash.

# Whats changing?
Add guards and a test for breaking the guards

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [x] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
+++ b/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
@@ -171,7 +171,11 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
 
     public void markElevatorAsCalibratedAgainstLowerLimit(){
         isCalibrated = true;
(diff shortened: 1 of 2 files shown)


src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -171,7 +171,11 @@ public void setPower(double power) {
 
     public void markElevatorAsCalibratedAgainstLowerLimit(){
         isCalibrated = true;
-        elevatorPositionOffset = this.masterMotor.getPosition().in(Rotations);
+        if (this.masterMotor != null) {
```
aschokking, line 174: ⭐ so far we've been having the students check the electricalContract here instead of the motor being null or not, so might want to keep it consistent until/if we switch over

aschokking (approved):

PR #124, opened 2025-02-08
Dont allow calibration without sensor

junyu101:
# Why are we doing this?
Theres no sensor on the robot right now, so it continuously applies downforce to try and calibrate even when its at the bottom, this is to prevent that for testing pursposes.
Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
+++ b/src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
@@ -97,23 +97,21 @@ public class ElevatorMaintainerCommand extends BaseMaintainerCommand<Distance> {
                 XCANMotorController.MotorPidMode.Voltage);
     }
 
+    //defaults humanControlAction if there is no bottom sensor
     @Override
     protected void uncalibratedMachineControlAction() {
-        var mode = calibrationDecider.decideMode(elevator.isCalibrated());
-
-        if(!contract.isElevatorBottomSensorReady()){
-            mode = CalibrationDecider.CalibrationMode.GaveUp;
-        }
+        var mode = contract.isElevatorBottomSensorReady() ?
+                calibrationDecider.decideMode(elevator.isCalibrated()) : CalibrationDecider.CalibrationMode.GaveUp;
 
         switch (mode){
             case Calibrated -> calibratedMachineControlAction();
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -98,13 +103,16 @@ protected void uncalibratedMachineControlAction() {
 
         switch (mode){
             case Calibrated -> calibratedMachineControlAction();
-            case Attempting -> attemptCalibration();
+            case Attempting -> attemptCalibrationIfSensorExists();
```
aschokking, line 106: ⭐ ⭐ ⭐  different idea, if the sensor doesn't exist have the mode just start in GaveUp? That way in starts in human control action right away

aschokking (changes requested):

JohnGilb (approved):

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -96,15 +101,19 @@ protected void calibratedMachineControlAction() {
     protected void uncalibratedMachineControlAction() {
         var mode = calibrationDecider.decideMode(elevator.isCalibrated());
```
aschokking, line 102: ⭐ would put this in the else of the `if` so it's less confusing

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -96,15 +101,19 @@ protected void calibratedMachineControlAction() {
...
             case Calibrated -> calibratedMachineControlAction();
-            case Attempting -> attemptCalibration();
+            case Attempting -> attemptCalibrationIfSensorExists();
             case GaveUp -> humanControlAction();
             default -> humanControlAction();
         }
     }
 
-    private void attemptCalibration(){
+    private void attemptCalibrationIfSensorExists(){
```
aschokking, line 116: ⭐ undo renaming this if we don't check sensor in here anymore

aschokking (approved):

PR #130, opened 2025-02-09
Safely testing superstructure, Set up buttons for coral and elevator on gamepad, and lowered human power limits

junyu101:
# Why are we doing this?
To test the superstructure on the real robot
Asana task URL:

# Whats changing?
Added calibration toggles for both arm and elevator, removed arm starting default calibrataed to true

# How this was tested
- [ ] tested on robot
- [x ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/coral_arm/commands/ToggleCoralArmCalibratedCommand.java
```diff
@@ -0,0 +1,29 @@
+package competition.subsystems.coral_arm.commands;
+
+import competition.subsystems.coral_arm.CoralArmSubsystem;
+import xbot.common.command.BaseCommand;
+
+import javax.inject.Inject;
+
+//toggles isCalibrated on and off in coralArmSubsystem for testing
+public class ToggleCoralArmCalibratedCommand extends BaseCommand {
```
stephenjust, line 9: protip: You can use an InstantCommand directly in the OperatorCommandMap instead of specifying a whole command here.

stephenjust (approved):

src/main/java/competition/subsystems/coral_arm/commands/ToggleCoralArmCalibratedCommand.java
aschokking: I will say, I think we abused InstantCommands in OperatorCommandMap in the past, making it so the file was enormous and hard to find where commands were actually implemented so I don't mind having classes (or maybe factory methods on the relevant subsystem?)

PR #132, opened 2025-02-09
Superstructure Bringup

JohnGilb:
# Why are we doing this?
These are the set of changes needed to run the 2025 superstructure using the Robox. 
Asana task URL:

# Whats changing?
- Lots of current limits, ID changes
- Fix a lot of incorrect prefixes on devices (makes them harder to organize in AdvantageScope)
- Fixed bugs relating to rotations per degrees and degrees per rotations - they need to be updated simultaneously.
- Adds force calibrate to arm
- 

# How this was tested
- [x] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -165,6 +174,9 @@ public void setPower(double power) {
             if (!isCalibrated){
                 power = MathUtils.constrainDouble(power,calibrationNegativePower.get(),0);
             }
+
+            aKitLog.record("ElevatorPower", power);
```
aschokking, line 178: don't we get this from the akit logging of the motors inputs?

aschokking (approved):

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
JohnGilb: Yeah, good catch. This is leftover from trying to debug an issue where the elevator wasn't moving and I started checking increasingly unlikely places things could go wrong.

PR #134, opened 2025-02-11
CoralArmSim: Fix for changes made for real robot

aschokking:
# Why are we doing this?

John hacked away the coral arm working in angles and instead it works in rotations right now. This totally borked the sim logic. This PR makes the arm work in the sim (though the robots sense of where the arm is is still kinda messed up).

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/coral_arm/CoralArmSubsystem.java
+++ b/src/main/java/competition/subsystems/coral_arm/CoralArmSubsystem.java
@@ -85,7 +85,7 @@ public class CoralArmSubsystem extends BaseSetpointSubsystem<Angle> {
             this.lowSensor = null;
         }
 
-        this.degreesPerRotations = propertyFactory.createPersistentProperty("Degrees Per Rotations", 0.1);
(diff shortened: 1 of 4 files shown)


src/main/java/competition/subsystems/coral_arm/CoralArmSubsystem.java
```diff
@@ -85,7 +85,7 @@ public CoralArmSubsystem(XCANMotorController.XCANMotorControllerFactory xcanMoto
             this.lowSensor = null;
         }
 
-        this.degreesPerRotations = propertyFactory.createPersistentProperty("Degrees Per Rotations", 0.1);
+        this.degreesPerRotations = propertyFactory.createPersistentProperty("Degrees Per Rotations", 6.94444);
```
aschokking, line 88: This is close to what the true value should be but we need to do a better measurement

PR #139, opened 2025-02-14
Terminal approach with BOTH front AND back cameras

Rongrrz:
# Why are we doing this?
A more dynamic terminal approach

Asana task URL:

# Whats changing?

I didn't delete any old things, only added new things

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -49,11 +52,23 @@ public Translation2d getReefAprilTagCameraData() {
         return new Translation2d(data.getX(), data.getY());
     }
 
+    public Translation2d getAprilTagCameraData(int cameraToUse) {
+        Transform3d data = getLatestTargetObservation(cameraToUse).cameraToTarget();
```
stephenjust, line 56: Try rotating this value by the Rotation3d of the camera.

src/main/java/competition/subsystems/drive/commands/AlignToTagGlobalMovementWithCalculator.java
```diff
@@ -0,0 +1,91 @@
...
+    public void setIsCameraBackwards(boolean backwards) {
+        this.backwards = backwards;
+    }
+
+    double headingAtCommandStart = 0;
+
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+        drive.getPositionalPid().reset();
```
JohnGilb, line 66: I suggest to move as much logic as possible into the calculator, such that this command is a very thin "shell" around that logic. 

This will be valuable when we want to have more kinds of combined drive, such as combining SwerveDriveWithJoysticks logic with this logic, allowing the driver to keep driving normally until the correct AprilTag is seen, at which point this new logic would take over.

So move all the PID resetting, initial heading detection into the calculator class. You can expose an `initialize` method there.

src/main/java/competition/subsystems/drive/commands/AlignToTagGlobalMovementWithCalculator.java
```diff
@@ -0,0 +1,91 @@
...
+        headingAtCommandStart = pose.getCurrentPose2d().getRotation().getDegrees();
+        calculator = new AlignCameraToAprilTagCalculator(aprilTagVisionSubsystem, drive.getPositionalPid(),
+                electricalContract, targetAprilTagID, targetCameraID, 24, backwards);
+    }
+
+    @Override
+    public void execute() {
+        Translation2d driveValues = calculator.getXYPowersAlignToAprilTag(pose.getCurrentPose2d());
```
JohnGilb, line 74: Similarly here - have the calculator return a Pose2d, with the translation representing x/y powers (as its doing already) and a rotation power value smuggled in via the "heading" of the Pose2d.

src/main/java/competition/subsystems/drive/commands/AlignToTagGlobalMovementWithCalculator.java
```diff
@@ -0,0 +1,91 @@
...
+        drive.fieldOrientedDrive(
+                new XYPair(driveValues.getX(), driveValues.getY()),
+                omega,
+                pose.getCurrentHeading().getDegrees(),
+                true);
+    }
+
+    @Override
+    public boolean isFinished() {
+        return drive.getPositionalPid().isOnTarget()
```
JohnGilb, line 87: similarly, pile this logic into the calculator, and have it expose a `recommendIsFinished()` method.

src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
```diff
@@ -0,0 +1,99 @@
...
+    private final boolean backwards;
+    AKitLogger akitLog;
+
+    public enum TagAcquisitionState {
+        NeverSeen,
+        LockedOn,
+        Lost
+    }
+    private TagAcquisitionState tagAcquisitionState = TagAcquisitionState.NeverSeen;
+
```
JohnGilb, line 34: nit: The constructor should be the first "function" in the class for readability.

src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
```diff
@@ -0,0 +1,99 @@
...
+        this.backwards = backwards;
+
+        this.alignmentPointOffset = generateAlignmentPointOffset(
+                electricalContract.getDistanceFromCenterToOuterBumperX(),
+                electricalContract.getCameraInfo()[targetCameraID],
+                Inches.of(offsetInInches),
+                backwards
+        );
+
+        this.akitLog = new AKitLogger("AlignToTagGlobalMovementWithCalculator/");
```
JohnGilb, line 62: This logic may be used by multiple commands - but no more than one instance should probably be running at a time. Either change this string to `AlignCameraToAprilTagCalculator` or take in a prefix from the parent that created you.

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -49,11 +54,28 @@ public Translation2d getReefAprilTagCameraData() {
         return new Translation2d(data.getX(), data.getY());
     }
 
+    public Translation2d getAprilTagCameraData(int cameraToUse) {
```
JohnGilb, line 57: The method name suggests it will get data from or about the AprilTagCamera, but the return type is a location. Where possible, we should be very clear about what a function will do since there's lots of subtle differences in the kinds of information we need and what photonvision can provide. Based on the logic of this method, this would be something like:

`GetRobotRelativeLocationOfBestDetectedAprilTagFromSpecificCamera`

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -49,11 +54,28 @@ public Translation2d getReefAprilTagCameraData() {
...
+        );
+
+        return new Translation2d(data.getX(), data.getY());
+    }
+
+    public double getAprilTagFieldOrientedRotation(int targetAprilTagID) {
+        return aprilTagFieldLayout.getTagPose(targetAprilTagID).get().getRotation().getZ();
+    }
+
     public boolean reefAprilTagCameraHasCorrectTarget(int targetAprilTagID) {
```
JohnGilb, line 69: not code in your PR, but another change here to update an ambiguous name. This is something like:
`doesBestObservationHaveAprilTagId()`

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -49,11 +54,28 @@ public Translation2d getReefAprilTagCameraData() {
...
         AprilTagVisionIO.TargetObservation targetObservation = getLatestTargetObservation(0);
         return targetObservation.fiducialId() == targetAprilTagID;
     }
 
+    public boolean aprilTagCameraHasCorrectTarget(int targetAprilTagID, int cameraToUse) {
```
JohnGilb, line 74: This method is in your PR, though, and would suggest rename:

`doesCameraBestObservationHaveAprilTagId(int cameraToUse, int targetAprilTagID)`

src/test/java/competition/subsystems/drive/DriveSubsystemTest.java
```diff
@@ -12,30 +11,6 @@ public class DriveSubsystemTest extends BaseCompetitionTest {
     @Test
     public void testDriveSubsystem() {
         DriveSubsystem driveSubsystem = (DriveSubsystem)getInjectorComponent().driveSubsystem();
-        assertNotNull(driveSubsystem);
```
JohnGilb, line 15: Was this intentional, or part of some merge?

JohnGilb (changes requested):
See my comments (accidentally did them as one-offs rather than a complete review)

src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
aschokking: I thought static stuff often came before the constructor since it was kind of meta to the class? I've at least since this style convention pitched before.

src/test/java/competition/subsystems/vision/AlignCameraToAprilTagCalculatorTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Test;
+import xbot.common.injection.electrical_contract.CameraInfo;
+import xbot.common.subsystems.vision.CameraCapabilities;
+
+import java.util.EnumSet;
+
+public class AlignCameraToAprilTagCalculatorTest extends BaseCompetitionTest {
+
+    @Test
+    public void testAlignmentPointOffset() {
```
aschokking, line 25: ⭐ for tricky math tests like this, I think it can help future generations to include more of a comment at the top here about what's being tested, and/or have comments in the test body before each assert about why things should be what they are

aschokking (approved):
Great work Rong!

src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
JohnGilb: Missed the static, I can support that!

JohnGilb (approved):
Approving as this is good progress. Take note of the 2 & 3 star comments.

src/test/java/competition/subsystems/drive/DriveSubsystemTest.java
Rongrrz: Friendly fire 😅

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
Rongrrz: I'll rename it in this PR, and in a separate PR once this one gets merged, I'll remove the older logic.

src/main/java/competition/subsystems/drive/commands/AlignToTagGlobalMovementWithCalculator.java
```diff
@@ -0,0 +1,91 @@
...
+        headingAtCommandStart = pose.getCurrentPose2d().getRotation().getDegrees();
+        calculator = new AlignCameraToAprilTagCalculator(aprilTagVisionSubsystem, drive.getPositionalPid(),
+                electricalContract, targetAprilTagID, targetCameraID, 24, backwards);
+    }
+
+    @Override
+    public void execute() {
+        Translation2d driveValues = calculator.getXYPowersAlignToAprilTag(pose.getCurrentPose2d());
+        aKitLog.record("driveValues", driveValues);
+        double rotation = aprilTagVisionSubsystem.getAprilTagFieldOrientedRotation(targetAprilTagID);
```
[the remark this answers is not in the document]
Rongrrz: Fixed!

src/main/java/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -60,26 +60,38 @@ public void setupDriverCommands(
...
+//        // operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.LeftBumper).whileTrue(scoreCoralCommand);
+        var alignToTag1 = alignToTag.get();
+        alignToTag1.setConfigurations(0, 18, false, 12);
+
+        var alignToTag2 = alignToTag.get();
+        alignToTag2.setConfigurations(1, 18, false, 12);
+
+        var alignToTag3 = alignToTag.get();
+        alignToTag3.setConfigurations(2, 13, true, 12);
+        operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.A).whileTrue(alignToTag1);
```
aschokking: ⭐ ⭐ Make sure @kujo27 approves the state of the driverGamepad here after these changes

aschokking (approved):
Changes from before look reasonable to me

aschokking (approved):
Also for things like this, please include videos of the simulator working to show how you tested your code

PR #142, opened 2025-02-15
First Generation of "Queue" control scheme using NeoTrellis

JohnGilb:
# Why are we doing this?
The oracle can follow a set of queued tasks, but there's no system to enqueue those tasks. For example, before the match starts, the drive team will want to select our coral scoring locations.
Asana task URL:

# Whats changing?
- NeoTrellis now parsed for combo actions
- Top 5 queued actions can be displayed on a dashboard

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [x] unit tests added

-----

PR feedback legend

 

src/main/java/competition/operator_interface/NeoTrellisSubsystem.java
```diff
@@ -0,0 +1,212 @@
...
+public class NeoTrellisSubsystem extends BaseSubsystem {
+
+    final OperatorInterface oi;
+    final XJoystick neoTrellis;
+    final ScoringQueue scoringQueue;
+
+    HashMap<FaceBranch, Integer> locationsToButtonIndices;
+    HashMap<Landmarks.CoralLevel, Integer> levelsToButtonIndices;
+
+    HashMap<FaceBranch, AdvancedTrigger> locationsToButtons;
```
aschokking, line 46: ⭐ nit, these should be final probably too?

src/main/java/competition/operator_interface/NeoTrellisSubsystem.java
```diff
@@ -0,0 +1,212 @@
...
+        locationsToButtons.put(faceBranch, neoTrellis.getifAvailable(buttonIndex));
+        locationsToButtonIndices.put(faceBranch, buttonIndex);
+    }
+
+    /**
+     * This method is invoked when at least one location button and one action button are pressed.
+     * It should only take action on the "rising edge" of such an event to debounce the buttons.
+     * @param e EdgeType (from the Latch)
+     */
+    private void queueAppropriateAction(Latch.EdgeType e) {
```
aschokking, line 116: point of discussion, but I'm not sure what we gain by making anything private in our codebase. If someone extended this class they should be able to call this still right?

aschokking (approved):

Rongrrz (commented):
Looks good to me, it seems a lot more dynamic now that it also has algae removal tasks and also being able to add tasks with the NeoTrellis (which I am much more unfamiliar with).

PR #144, opened 2025-02-15
Reset pose estimators on disabledInit()

stephenjust:
# Why are we doing this?
Vision team is observing some weirdness when we lose track of vision pose. We can reduce the number of possible failure points by resetting the pose estimations using the vision pose on init.

# Whats changing?
Add some helpers commands to reset the pose estimators and clear out state.
On disabledInit() the pose estimators will reset to the pose reported by the vision estimator.

We might want to run this at other times too, but this is just a starting point for easy testing.

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/Robot.java
```diff
@@ -107,6 +107,8 @@ public BaseRobotComponent getInjectorComponent() {
     public void disabledInit() {
         super.disabledInit();
         reachedDisabledInit.countDown();
+        var poseSub = (PoseSubsystem) getInjectorComponent().poseSubsystem();
```
aschokking, line 110: ⭐ I don't love this casting, maybe we should save a ref to PoseSubsystem in the class so we only need to do it once when we create the subsystem? Or does Dagger have a better way of having this be strongly typed?

src/main/java/competition/Robot.java
```diff
@@ -107,6 +107,8 @@ public BaseRobotComponent getInjectorComponent() {
...
+        poseSub.getResetTranslationToVisionEstimateCommand().schedule();
```
aschokking, line 111: ❓ I was curious, why do this as a command that gets scheduled vs doing the reset directly on the pose subsystem in this moment?

aschokking (approved):

src/main/java/competition/Robot.java
stephenjust: I did it because I was lazy and only wanted to implement it once, but there are benefits to doing it this way - this guarantees it runs after refreshDataFrame() and periodic() so there will be some vision state available even on a fresh boot.

src/main/java/competition/Robot.java
stephenjust: Fixed.

PR #150, opened 2025-02-15
Enum for cameras

Rongrrz:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
@@ -2,104 +2,54 @@ package competition.subsystems.drive.commands;
 
 import competition.electrical_contract.ElectricalContract;
 import competition.subsystems.drive.DriveSubsystem;
-import competition.subsystems.pose.Landmarks;
 import competition.subsystems.pose.PoseSubsystem;
 import competition.subsystems.vision.AprilTagVisionSubsystemExtended;
 import edu.wpi.first.math.geometry.Pose2d;
-import edu.wpi.first.math.geometry.Pose3d;
-import edu.wpi.first.math.geometry.Rotation3d;
-import edu.wpi.first.math.geometry.Translation2d;
-import edu.wpi.first.units.measure.Distance;
-import edu.wpi.first.wpilibj.DriverStation;
-import xbot.common.command.BaseCommand;
-import xbot.common.injection.electrical_contract.CameraInfo;
-import xbot.common.math.XYPair;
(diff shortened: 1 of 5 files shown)


src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
```diff
@@ -2,104 +2,54 @@
...
-        targetReefFacePose = pose.getClosestReefFacePose();
-        aKitLog.record("TargetReefFace", targetReefFacePose);
+        Pose2d targetReefFacePose = pose.getClosestReefFacePose();
+        super.setConfigurations(
+                cameraToUse,
+                aprilTagVisionSubsystem.getTargetAprilTagID(targetReefFacePose),
+                isCameraBackwards,
+                offsetInInches
+        );
+        super.initialize();
```
stephenjust, line 43: ⭐Conventionally, you should call the `super` method first before you do anything else.

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
```diff
@@ -2,104 +2,54 @@
...
-        double omega = headingModule.calculateHeadingPower(targetReefFacePose.getRotation().getDegrees());
-
-        aKitLog.record("PowerVector", driveValues);
-        aKitLog.record("omega", omega);
-        drive.move(new XYPair(driveValues.getX(), driveValues.getY()), omega);
+        super.execute();
     }
 
     @Override
     public boolean isFinished() {
```
stephenjust, line 52: ⭐⭐Since isFinished only calls the `super` method now, you can remove it entirely.

stephenjust (approved):

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
Rongrrz: The awkward thing is the calculator gets instantiated in super.initialize() and I need the configurations to be set before it, but also just right before it so that the robot's pose accurately reflects the current pose 😅

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
Rongrrz: I'll remove execute as well :)

PR #151, opened 2025-02-15
Adding vibrations

Glowdisk:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [x] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/coral_scorer/commands/IntakeCoralCommand.java
```diff
@@ -3,21 +3,33 @@
...
+
     @Override
     public void initialize() {
         coral.intake();
     }
 
+    @Override
+    public void execute(){
+        if (coral.confidentlyHasCoral()){
+            rumble.rumbleGamepad(100,100);
```
aschokking, line 30: ⭐ ⭐ ⭐ This is going to crash because rumble is `null` AND this isn't actually how you get a rumble manager.

 Here's some example code from 2024 you can use as an example for how to set the rumble correctly: `oi.driverGamepad.getRumbleManager().rumbleGamepad(higherIntensity.get(), 0.7);`

src/main/java/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -77,7 +80,7 @@ public void setupDriverCommands(
         var homed = prepCoralSystemCommandGroupFactory.create(Landmarks.CoralLevel.COLLECTING);
         operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.B).onTrue(homed);
 
-        operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.RightBumper).whileTrue(intakeCoralCommand);
+        operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.RightBumper).whileFalse(intakeCoralCommand);
```
aschokking, line 83: ⭐ ⭐ ⭐ don't change this to whileFalse, it would mean the robot would always be trying to intake coral all the time

aschokking (changes requested):

PR #159, opened 2025-02-16
Adding a drive to coral station via vision avoidance system.

rokadias:
# Why are we doing this?

Asana task URL:

# Whats changing?
Adding the ability to use Vision path planning to avoid obstacles for designated locations.

Currently set to the only april tag on our setup field, will update to any coral station based on location and which side of the field we are operating on.

Still needs the corresponding code on Alt side.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [x] unit tests added

-----

PR feedback legend

 

JohnGilb (approved):
lgtm!

PR #165, opened 2025-02-17
Trapezoid profile manager: store prev setpoint internally 

aschokking:
# Why are we doing this?

and only use it on reset if it's been recently used.

This fixes the bug where after disabling the robot and the elevator drops down, when you renable it it was using the previous setpoint to start the profile.

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/motion/TrapezoidProfileManager.java
+++ b/src/main/java/competition/motion/TrapezoidProfileManager.java
@@ -33,9 +33,9 @@ public class TrapezoidProfileManager {
     TrapezoidProfile.Constraints constraints;
     TrapezoidProfile.State initialState;
     TrapezoidProfile.State goalState;
-    MutTime profileStartTime = Seconds.of(-10000).mutableCopy();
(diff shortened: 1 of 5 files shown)


src/main/java/competition/motion/TrapezoidProfileManager.java
```diff
@@ -29,7 +33,9 @@ public abstract TrapezoidProfileManager create(
     TrapezoidProfile.Constraints constraints;
     TrapezoidProfile.State initialState;
     TrapezoidProfile.State goalState;
-    Time profileStartTime = Seconds.zero(); 
+    MutTime profileStartTime = Seconds.of(-10000).mutableCopy();
```
stephenjust, line 36: ⭐⭐Use Seconds.mutable to start with a mutable object.

stephenjust (approved):

PR #174, opened 2025-02-19
Will stalk april tag until it gets near (terminal approach fix)

Rongrrz:
# Why are we doing this?
So we don't change our heading and all of a sudden we lose our tag target

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

https://github.com/user-attachments/assets/6f0ce98c-0848-4194-92e8-56315385c312

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
+++ b/src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
@@ -9,6 +9,7 @@ import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.geometry.Rotation3d;
 import edu.wpi.first.math.geometry.Transform2d;
 import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.units.measure.Angle;
 import edu.wpi.first.units.measure.Distance;
 import xbot.common.advantage.AKitLogger;
 import xbot.common.injection.electrical_contract.CameraInfo;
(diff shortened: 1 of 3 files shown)


src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
```diff
@@ -38,6 +39,8 @@ public enum TagAcquisitionState {
     final Rotation3d cameraRotation;
     Translation2d targetLocationOnField = new Translation2d(0, 0);
 
+    final double distanceForExactAlignment = 1;
```
JohnGilb, line 42: 1 what? meters? (it's okay to have a double here, just make sure the name is descriptive)

src/main/java/competition/subsystems/vision/AlignCameraToAprilTagCalculator.java
```diff
@@ -104,20 +108,32 @@ public Pose2d getXYPowersAlignToAprilTag(Pose2d currentPose) {
             case LockedOn, Lost -> {
                 // Flip our desiredRotation if backwards
                 Optional<Pose3d> aprilTagPose = aprilTagVisionSubsystem.getAprilTagFieldOrientedPose(targetAprilTagID);
-                double desiredRotation = aprilTagPose.map(
-                        (tag) -> Math.PI + tag.getRotation().getZ() - cameraRotation.getZ()
-                ).orElse(initialHeading);
-                akitLog.record("desiredRotation", desiredRotation);
+                double desiredHeading = initialHeading; // Or should we use current heading to avoid possible oscillation?
+                if (aprilTagPose.isPresent()) {
```
JohnGilb, line 112: I think you should always be able to point at the apriltag, even if you lose it. We still know where it "should" be given our location on the field.

JohnGilb (approved):

PR #179, opened 2025-02-21
Attempt to fix elevator jerkiness on enable/disable

stephenjust:
# Why are we doing this?
The elevator has some unexpected motion when enabling after disable.

# Whats changing?
Explicitly reset the state on initializing the maintainer or starting machine control.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [x] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/motion/TrapezoidProfileManager.java
+++ b/src/main/java/competition/motion/TrapezoidProfileManager.java
@@ -55,6 +55,13 @@ public class TrapezoidProfileManager {
         goalState = new TrapezoidProfile.State(initialPosition, 0);
     }
 
+    public void resetState(double currentValue, double currentVelocity) {
+        initialState = new TrapezoidProfile.State(currentValue, currentVelocity);
+        goalState = new TrapezoidProfile.State(currentValue, 0);
+        previousSetpoint = currentValue;
+        profileStartTime.mut_replace(XTimer.getFPGATimestampTime().minus(Seconds.of(0.02)));
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/competition/motion/TrapezoidProfileManager.java
```diff
@@ -55,6 +55,13 @@ public TrapezoidProfileManager(
         goalState = new TrapezoidProfile.State(initialPosition, 0);
     }
 
+    public void resetState(double currentValue, double currentVelocity) {
+        initialState = new TrapezoidProfile.State(currentValue, currentVelocity);
+        goalState = new TrapezoidProfile.State(currentValue, 0);
+        previousSetpoint = currentValue;
+        profileStartTime.mut_replace(XTimer.getFPGATimestampTime().minus(Seconds.of(0.02)));
```
aschokking, line 62: ⭐ ⭐  Should probably be setting the previousSetpointTime here too right?

aschokking (approved):
Would love to understand more about the enable cases this catches that the other logic was missing

src/main/java/competition/motion/TrapezoidProfileManager.java
stephenjust: Don't even know if this works yet. John broke the elevator before I could test this on the real bot.

JohnGilb: The elevator... broke itself. (Sprocket teeth wearing down). I'm only tangentially responsible!! :D

JohnGilb (approved):

PR #180, opened 2025-02-21
Feature/spi adapter

zzzoziea:
# Why are we doing this?

Asana task URL:

# Whats changing?
Added XSpiController and WpiLibSpiAdapter. Also updated Light Subsytem with SPI protocol to constantly send robot state over SPI

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/lights/LightSubsystem.java
+++ b/src/main/java/competition/subsystems/lights/LightSubsystem.java
@@ -2,6 +2,8 @@
 
 package competition.subsystems.lights;
 
+import java.nio.ByteBuffer;
+
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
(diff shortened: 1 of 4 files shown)


src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -87,9 +92,16 @@ public LightsStateMessage getCurrentState() {
...
-        // TODO: decide on how communication will actually happen
+    public void sendState(LightsStateMessage state) {        
+        try {
+            ByteBuffer lightsState = ByteBuffer.allocate(Integer.BYTES);
+            lightsState.putInt(state.getValue());
+
+            spiController.write(lightsState, Integer.BYTES);
+        } catch(Exception e) {
+            spiController.close();
+            throw e;
```
stephenjust, line 103: ⭐⭐⭐Robot code should not throw exceptions. Log the error and find a way to make the robot keep running!

stephenjust: Do we need to re-initialize the spiController after we close it?

stephenjust (changes requested):

src/main/java/competition/subsystems/lights/LightSubsystem.java
zzzoziea: mmm I think if we hit an exception it might be cooked but maybe we can reinit somehow? I'm not really sure how we handle recovery on the robot

JohnGilb: For situations like this we can catch the exception, log, and then set some boolean "broken" state where all further calls to the device are redirected to do nothing.

PR #181, opened 2025-02-21
Vibrations for the intake

Glowdisk:
There is a bug where you lift the trigger and it still rumbles so but im not sure is it because its still has coral

# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

JohnGilb (commented):
Still a mystery here why the gamepad never stops rumbling. We need to figure out why that's getting stuck.

src/main/java/competition/subsystems/coral_scorer/commands/IntakeCoralCommand.java
```diff
@@ -1,23 +1,55 @@
...
     }
 
+
+    @Override
+    public void execute(){
+        double triggerValue = oi.superstructureGamepad.getRawAxis(XboxController.Axis.kLeftTrigger.value);
+
+        if (coral.confidentlyHasCoral() && triggerValue > 0.5) {
+            System.out.println("Execute happened - Rumble On");
+            oi.superstructureGamepad.getRumbleManager().rumbleGamepad(operatorGamePadRumblePrefrence.get(), 1);
```
aschokking, line 40: ⭐ ⭐ ⭐ shouldn't this be operatorGamepad not superstructureGamepad? The latter is used for testing/debugging but won't be used by Anthony in a real match.

aschokking: That's also why the rumbling won't stop, superstructure gamepad doesn't have it's periodic being called (I only hooked up the driver and operator gamepads to that)

aschokking (changes requested):

src/main/java/competition/subsystems/coral_scorer/commands/IntakeCoralCommand.java
```diff
@@ -1,23 +1,55 @@
...
+
+    @Override
+    public void end(boolean interrupted) {
+        oi.superstructureGamepad.getRumbleManager().rumbleGamepad(0, 0);
+        oi.driverGamepad.getRumbleManager().rumbleGamepad(0.0, 0);
+    }
+
+    @Override
+    public boolean isFinished() {
+        return oi.superstructureGamepad.getRawAxis(XboxController.Axis.kLeftTrigger.value) <= 0.1;
```
aschokking, line 53: ⭐ ⭐ ⭐ this isFinished logic doesn't make a lot of sense to me, why would we be looking at the left trigger value being low?

aschokking (changes requested):

PR #190, opened 2025-02-22
Use XDutyCycleEncoder for the coral arm sensor

stephenjust:
XAbsoluteEncoder is not meant to handle anything but a cancoder, we were reading the wrong type of encoder.

# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -110,12 +110,12 @@ public CANMotorControllerInfo getAlgaeArmPivotMotor() {
     public boolean areCanCodersReady() {
         return true;
     }
-    // change channels
+
     public DeviceInfo getCoralArmPivotAbsoluteEncoder() {
-        return new DeviceInfo("ArmPivotAbsoluteEncoder", 100);
+        return new DeviceInfo("ArmPivotAbsoluteEncoder", 6);
```
stephenjust, line 115: Wait for confirmation from electrical on ID!

stephenjust: Electrical says 1 or 9. Try both before merging.

JohnGilb (approved):
Approved, but agreed should not merge until we can confirm IDs.

PR #194, opened 2025-02-23
Four coral auto left faces

Rongrrz:
# Why are we doing this?

Asana task URL:

# Whats changing?

@anthonytrenh I've started on converting landmarks blue to red if needed, since you are more familiar with the system, you should fill in the rest.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/auto_programs/FromLeftCageScoreFarLeftBranchALevelFour.java
```diff
@@ -0,0 +1,41 @@
...
+        this.autoSelector = autoSelector;
+
+        // Force our location to start in front of left cage
+        var startInFrontOfLeftCage = pose.createSetPositionCommand(
+                () -> PoseSubsystem.convertBlueToRedIfNeeded(Landmarks.BlueLeftStartingLine)
+        );
+        this.addCommands(startInFrontOfLeftCage);
+
+        // Drive to far left, branch A and score level four
+        queueMessageToAutoSelector("Drive to far left, branch A and score level four");
```
aschokking, line 33: ⭐ ⭐ could this message be procedurally generated from the target location and branch and level?

src/main/java/competition/auto_programs/FromLeftCageScoreFarLeftBranchALevelFour.java
```diff
@@ -0,0 +1,41 @@
...
+        var driveAndScoreFarLeftBranchALevelFour = driveToFaceAndScoreFact.create(Landmarks.ReefFace.FAR_LEFT, Landmarks.Branch.A, Landmarks.CoralLevel.FOUR);
+        this.addCommands(driveAndScoreFarLeftBranchALevelFour);
+    }
+
+    private void queueMessageToAutoSelector(String message) {
```
aschokking, line 38: ⭐ ⭐ we should really make a shared base class that has this on it so we don't keep copying and pasting it all over

src/main/java/competition/auto_programs/FromLeftCageScoreFarLeftBranchALevelFour.java
aschokking: My concern is it's really easy to make a mistake here and have this string not align with where the robot is actually going.

src/main/java/competition/auto_programs/FromLeftCageScoreFarLeftBranchALevelFour.java
```diff
@@ -0,0 +1,41 @@
...
+
+import javax.inject.Inject;
+import java.util.Set;
+
+public class FromLeftCageScoreFarLeftBranchALevelFour extends SequentialCommandGroup {
+
+    final AutonomousCommandSelector autoSelector;
+
+    @Inject
+    public FromLeftCageScoreFarLeftBranchALevelFour(AutonomousCommandSelector autoSelector,
```
aschokking, line 21: ⭐ ⭐ it seems like instead of having all these classes that are very similar, maybe a factory that creates these command groups given a starting location and target branch/level?

aschokking (approved):
Thanks for tackling this, looks like a great start! Made a few suggestions to help DRY the code up a bit.

aschokking (approved):
A ton cleaner than before

PR #198, opened 2025-02-26
When aligning to an April Tag, lock on to a single tag even if it's not the best target

stephenjust:
# Why are we doing this?
If the target tag is no longer the best tag, the robot could lose it. Updated the code to always track a specific tag.

# Whats changing?
Use the new commands on AprilTagVisionSubsystem to track a specific tag rather than the best.

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

https://github.com/user-attachments/assets/8f2eeabe-9d57-4026-9b04-867ff793a162

-----

PR feedback legend

 

Rongrrz (approved):
I see no problems, reasonable changes based on current logic

aschokking (approved):
makes sense!

PR #200, opened 2025-02-26
Added light enums

4fishs:
# Why are we doing this?
We need to add light enums for our light system
Asana task URL:

# Whats changing?
New light states

# How this was tested
- [X ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -27,12 +27,27 @@ public class LightSubsystem extends BaseSubsystem {
     public enum LightsStateMessage{
         // we never send NoCode, it's implicit when the robot is off
         // and all of the DIOs float high
-        NoCode(maxValue), 
+        NoCode(0),
         RobotDisabled(1),
         RobotEnabled(2),
-        CoralPresent(3),
-        RequestCoralFromHuman(4),
```
aschokking, line 34: ⭐ ⭐ ⭐ this old enum value was being referenced elsewhere in the code in this file, you'll have to update that reference to it to the new value or the code won't compile

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -27,12 +27,27 @@ public class LightSubsystem extends BaseSubsystem {
...
-        ReadyToScore(5);
```
aschokking, line 35: ⭐ ⭐ ⭐ this one too

aschokking (changes requested):

src/main/java/competition/subsystems/lights/LightSubsystem.java
4fishs: I just updated both lines! Can you check it out?

aschokking (approved):
Looks great to me, thanks for adding these!

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -82,7 +97,7 @@ public LightsStateMessage getCurrentState() {
         } else if (coralScorerSubsystem.confidentlyHasCoral()) {
             currentState = LightsStateMessage.CoralPresent;
         } else if (coralScorerSubsystem.getCoralScorerState() == CoralScorerSubsystem.CoralScorerState.INTAKING) {
-            currentState = LightsStateMessage.RequestCoralFromHuman;
+            currentState = LightsStateMessage.CoralReadyToReceive;
```
aschokking, line 100: ⭐ ⭐ ⭐ ah actually looks like this enum value doesn't exist so it won't compile

aschokking (changes requested):

aschokking (approved):
That did it!

PR #201, opened 2025-02-26
DriverRelativeBranchAlignment

kujo27:
# Why are we doing this?
Allows for easy branch alignment
Asana task URL:

# Whats changing?

OperatorCommandMap, AlignToReefWithAprilTagCommand, PoseSubsystem
# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

https://github.com/user-attachments/assets/147c664a-1043-4192-9859-1843879ae9eb

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
@@ -5,10 +5,11 @@ import competition.subsystems.drive.DriveSubsystem;
 import competition.subsystems.drive.logic.AlignCameraToAprilTagCalculator;
 import competition.subsystems.pose.PoseSubsystem;
 import competition.subsystems.vision.AprilTagVisionSubsystemExtended;
-import edu.wpi.first.math.geometry.Pose2d;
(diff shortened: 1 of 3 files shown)


stephenjust (approved):
Makes sense. Can you add a video from the simulator?

src/main/java/competition/subsystems/drive/commands/AlignToReefWithAprilTagCommand.java
```diff
@@ -35,12 +38,40 @@ public void setConfigurations(int cameraToUse, boolean isCameraBackwards, double
...
-        );
+            );
         super.initialize();
     }
+
+    public void setDriverRelative(boolean isEnabled) {
+        this.isDriverRelative = isEnabled;
+    }
+
+    private void setDriverRelativeCameraToUse() {
```
aschokking, line 58: ⭐ ⭐ This logic is really calling out for a unit test to verify the red/blue cases, even if it's working correctly now it would be easy for someone to break it when making updates here and not realize it without a lot of manual testing.

aschokking (approved):

PR #205, opened 2025-02-28
BlockVisionUpdatesWhenAligning

kujo27:
# Why are we doing this?
An approach that potentially makes the robot accurately align to reef tag.
Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -188,6 +188,7 @@ public AlignCameraToAprilTagAdvice getXYPowersAlignToAprilTag(Pose2d currentPose
         // Mostly, this is about where we should be pointing - and we generally point at the tag unless we are fairly close.
         Optional<AprilTagVisionIO.TargetObservation> targetObservation = aprilTagVisionSubsystem.getTargetObservation(targetCameraID, targetAprilTagID);
         boolean doWeSeeOurTargetTag = targetObservation.isPresent() && targetObservation.get().ambiguity() < maxTagAmbiguity.get();
+        pose.setVisionUpdatesDisabled(doWeSeeOurTargetTag);
```
aschokking, line 191: ⭐ I'm a little nervous about the disable being here but the enable being somewhere else. Could the command be doing both instead?

src/main/java/competition/subsystems/pose/PoseSubsystem.java
```diff
@@ -48,6 +47,8 @@ public class PoseSubsystem extends BasePoseSubsystem {
     public static final Distance fieldXMidpointInMeters = Meters.of(8.7785);
     public static final Distance fieldYMidpointInMeters = Meters.of(4.025);
 
+    private boolean isVisionUpdatesDisabled = false;
```
aschokking, line 50: ⭐ nit, since it's plural this should be `areVisionUpdatesDisabled` or maybe `isVisionUpdatingDisabled`

aschokking (approved):

PR #207, opened 2025-02-28
Trapezoid profile manager preventions

junyu101:
# Why are we doing this?
Preventing large accelerations by limiting profile computation to a tunable gap so that it only computes the rest of the profile when the robot "catches up"
Asana task URL:

# Whats changing?
Trapezoid profile manager

# How this was tested
- [ ] tested on robot
- [ x] tested in simulator
- [ ] unit tests added

(This is on the way up, the slow increase in position of the robot is because I tried turning off gravityfeedforward on the motor pid but it didn't do anything so I just tested it regardless, it does the same thing except with a slight upward movement trend)

image

(This is on the way down)

https://github.com/user-attachments/assets/69f85c96-5f55-4137-a87a-0fdfe5d03771

-----

PR feedback legend

 

src/test/java/competition/motion/TrapezoidProfileManagerTest.java
```diff
@@ -13,7 +13,7 @@ public class TrapezoidProfileManagerTest extends BaseCompetitionTest {
     public void setUp() {
         super.setUp();
         manager = getInjectorComponent().trapezoidProfileManagerFactory()
-                .create("test", 1, 1, 0);
+                .create("test", 1, 1, 1000, 0);
```
aschokking, line 16: ⭐ ⭐ bonus points for actually having the test verify the gap is respected

aschokking (approved):

PR #208, opened 2025-02-28
Feature/approach verifies accuracy

JohnGilb:
# Why are we doing this?
Sometimes, when the robot does vision approach, it knows it has a really bad offset but just goes ahead and does final approach. If our last view of the AprilTag showed us way offset, then we need to go back and try aligning again.
Asana task URL:

# Whats changing?
Adds an horizontal error margin, default of 2", for final alignment. Otherwise, it goes back to the interstitial point.

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -73,8 +73,9 @@ public enum Activity {
     final DoubleProperty shovePower;
     final DoubleProperty shoveDuration;
     final DoubleProperty maxTagAmbiguity;
+    final DoubleProperty maxHorizontalErrorMeters;
```
aschokking, line 76: ⭐ `DistanceProperty` is now available for this kind of thing if you want!

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -115,6 +116,7 @@ public AlignCameraToAprilTagCalculator(AprilTagVisionSubsystemExtended vision, D
         shovePower = pf.createPersistentProperty("ShovePower", 0.25);
         shoveDuration = pf.createPersistentProperty("ShoveDuration-s", 0.5);
         maxTagAmbiguity = pf.createPersistentProperty("MaxTagAmbiguity", 0.5);
+        maxHorizontalErrorMeters = pf.createPersistentProperty("MaxHorizontalError-m", 0.0508); // 2 inches
```
aschokking, line 119: If you used DistanceProperty you could just set this in inches like you wanted 😛

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -261,8 +263,14 @@ public AlignCameraToAprilTagAdvice getXYPowersAlignToAprilTag(Pose2d currentPose
 
                 // If we're quite close to the final point, advance to shoving into the reef or coral station.
                 if (currentPose.getTranslation().getDistance(targetLocationOnField) < distanceToStartShoving.get()) {
-                    activity = Activity.Shove;
-                    shoveStartTime = XTimer.getFPGATimestamp();
+                    // We need to make a decision. If our error is small enough, we should advance to shove.
+                    // However, if our error is large, we should retreat and try again.
+                    if (isLastKnownErrorWithinBounds()) {
+                        activity = Activity.Shove;
+                        shoveStartTime = XTimer.getFPGATimestamp();
```
aschokking, line 270: Just thinking out loud, I wonder if we should have a more generic state machine framework that on transition just starts a clock for `stateStartTime` that could be used for these kinds of common situations/timeouts?

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -289,13 +297,29 @@ public AlignCameraToAprilTagAdvice getXYPowersAlignToAprilTag(Pose2d currentPose
...
         return new AlignCameraToAprilTagAdvice(driveIntent, rotationIntent, tagAcquisitionState, activity);
     }
 
+    private boolean isLastKnownErrorWithinBounds() {
+        // We have to at least be in an Approach/Shove phase for this to be possibly true.
+        // If we're in the searching phase, error isn't relevant, so we return false.
+        return switch (activity) {
+            case ApproachWhileCentering, TerminalApproach, Shove ->
+                Math.abs(lastKnownHorizontalErrorMeters) < maxHorizontalErrorMeters.get();
+            default -> false;
```
aschokking, line 311: ⭐ this works, but I wonder if we should instead list out the irrelevant states explicitly, that way if a new state gets added later someone has to make an explicit choice here about where it belongs vs forgetting to update this spot?

aschokking (approved):
Bonus points for adding a video in the sim of this working

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
JohnGilb: we're still forced per language settings to include a default (maybe from checkstyle?), so that won't catch anything in the IDE or at compile time.

PR #209, opened 2025-02-28
One coral autos

anthonytrenh:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

anthonytrenh:
ScoreWhenReady not working in sim, but everything else does

src/main/java/competition/auto_programs/FromCageScoreOneCoralAutoFactory.java
```diff
@@ -12,14 +14,17 @@
...
     final AutonomousCommandSelector autoSelector;
+    MapleSimulator mapleSimulator;
     PoseSubsystem pose;
     DriveToFaceAndScoreCommandGroupFactory driveToFaceAndScoreCommandGroupFact;
 
     @Inject
     public FromCageScoreOneCoralAutoFactory(AutonomousCommandSelector autoSelector,
-                                            DriveToFaceAndScoreCommandGroupFactory driveToFaceAndScoreCommandGroupFact,
-                                            PoseSubsystem pose){
+                                            MapleSimulator mapleSimulator,
```
aschokking, line 23: ⭐ ⭐ ⭐ you 100% can't do this on the real robot, it will cause the simulator to be created and crash the robot as it assumes all the motors are Mock (when they're real). 

Instead you should ask for a `BaseSimulator` which will give you the MapleSim when simulating and the NoopSim on the real robot (that doesn't do anything at all. Or since this is a command group anyway, there's a command you can safely use that's doing this already: `ResetSimulatedPose`

aschokking (changes requested):

aschokking (approved):

PR #212, opened 2025-02-28
Some cleanup and the ability to drive to the closest reef section.

rokadias:
# Why are we doing this?

* Feedback from the last PR.
* Improvements from testing (whileTrue).
* Adding April tags for the different reefs
* Adding in flow that does the same for the closest reefs as a starting point.

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

drive-to-reef

-----

PR feedback legend

 

src/main/java/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -103,7 +105,8 @@ public void setupDriverCommands(
         var oracleControlsRobot = Commands.parallel(driveAccordingToOracle, superstructureAccordingToOracle);
 
         var homed = prepCoralSystemCommandGroupFactory.create(() -> Landmarks.CoralLevel.COLLECTING);
-        operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.Y).onTrue(driveToCoralStationWithVisionCommand);
+        operatorInterface.driverGamepad.getifAvailable(XXboxController.XboxButton.X).whileTrue(driveToClosestReefSectionWithVisionCommand);
```
aschokking, line 108: Want to make sure driver @kujo27 reviews changes to his gamepad button mapping

src/main/java/competition/subsystems/drive/commands/DriveToBezierCurvesWithVisionCommand.java
```diff
@@ -28,20 +30,24 @@
...
                                          PropertyFactory pf, HeadingModule.HeadingModuleFactory headingModuleFactory,
-                                         RobotAssertionManager assertionManager) {
+                                         RobotAssertionManager assertionManager, ElectricalContract electricalContract) {
         super(drive, pose, pf, headingModuleFactory, assertionManager);
         this.pose = pose;
         this.drive = drive;
         this.coprocessorComms = coprocessorComms;
+
+        var distanceToOuterBumerInMeters = electricalContract.getDistanceFromCenterToOuterBumperX().in(Units.Meters);
+        this.radiusOfRobot = Math.sqrt(Math.pow(distanceToOuterBumerInMeters, 2.0) * 2.0);
```
aschokking, line 50: ⭐ maybe something that should go on ElectricalContract as well if it might be useful in other spots?

src/main/java/competition/subsystems/drive/commands/DriveToClosestReefSectionWithVisionCommand.java
```diff
@@ -0,0 +1,64 @@
...
+    DriveToClosestReefSectionWithVisionCommand(PoseSubsystem pose, DriveSubsystem drive,
+            CoprocessorCommunicationSubsystem coprocessorComms,
+            PropertyFactory pf, HeadingModule.HeadingModuleFactory headingModuleFactory,
+            RobotAssertionManager assertionManager, ElectricalContract electricalContract,
+            AprilTagFieldLayout aprilTagFieldLayout) {
+        super(pose, drive, coprocessorComms, pf, headingModuleFactory, assertionManager, electricalContract);
+        this.aprilTagFieldLayout = aprilTagFieldLayout;
+    }
+
+    private Pose2d getClosestReefPose() {
```
aschokking, line 36: ⭐ ⭐ There's already a utility for doing this being used by the current autonomous logic, maybe it could be re-used here? 

https://github.com/Team488/TeamXbot2025/blob/main/src/main/java/competition/subsystems/pose/PoseSubsystem.java#L269

src/main/java/competition/subsystems/drive/commands/DriveToClosestReefSectionWithVisionCommand.java
```diff
@@ -0,0 +1,64 @@
...
+
+        var robotPose = this.pose.getCurrentPose2d();
+        return robotPose.nearest(reefPoses);
+    }
+
+    @Override
+    public void initialize() {
+        if (this.setDestinationPoseForVision(this.getClosestReefPose(), false)) {
+            super.initialize();
+        }
```
aschokking, line 62: ⭐ ⭐ might be a good idea to log if this initialize ever didn't fire about what happened?

aschokking (approved):

src/main/java/competition/subsystems/drive/commands/DriveToClosestReefSectionWithVisionCommand.java
rokadias: I don't really trust the values from the Landmarks yet, as the math for how those points were figured out is unclear to me.

They are supposed to already have the robot geometry to be accounted for, but it's not really shown. What I'm proposing here just uses the values from the april field layout and should work if it's Andymark or Welded. Also doesn't assume needing to translate based on red vs blue since it's much more fixed.

I think if these values work (and it does in sim, and looks about right when testing with our made up field). I'd likely suggest we transition to this compared to what's there, but I don't want to do that without ample testing.

src/main/java/competition/subsystems/drive/commands/DriveToClosestReefSectionWithVisionCommand.java
rokadias: This is really good idea, thanks!

src/main/java/competition/subsystems/drive/commands/DriveToBezierCurvesWithVisionCommand.java
rokadias: Sure, why not, and it's pretty easy to add.

src/main/java/competition/subsystems/drive/commands/DriveToClosestReefSectionWithVisionCommand.java
JohnGilb: :star: If you're concerned about ample testing, then I'd suggest using the library function, which has been tested throughout the season. If you believe this approach is better, then I'd recommend replacing the library function immediately so it can get as much testing as possibly as quickly as possible from all other commands being run. Either way, let's get out of the "2x ways to do things" state quickly.

rokadias: I agree with the intent of having a single approach. What I mentioned above is why I don't trust this current approach and the level of testing is unclear to me (see mention of Andymark vs Welded fields which would definitely be off by a few cm). Also assuming a mirroring translation is close enough for Red vs Blue.

See location differences here which can impact testing in practice fields vs the real field:
image
 
The idea here is to move a quickly here to have ample testing on a different approach and these are just steps towards that end goal of having a single approach. I need to build on top of this to verify it works in a number of cases. In order to do that we need to merge with other automation flows. I've tested the April tag approach with Coral Stations with the 2024 robot in the lobby, and a bunch in simulation.

src/main/java/competition/operator_interface/OperatorCommandMap.java
kujo27: I want it drive to closest reef section to be on Y

kujo27 (approved):

src/main/java/competition/operator_interface/OperatorCommandMap.java
rokadias: Swapped the X and the Y.

PR #214, opened 2025-03-01
New climber logic

junyu101:
# Why are we doing this?

Asana task URL:

# Whats changing?

Everything is done in a command instead of the subsystem is that going to be a problem?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -349,4 +349,19 @@ public CANMotorControllerInfo getClimberMotor() {
                 new CANMotorControllerOutputConfig()
                         .withStatorCurrentLimit(Amps.of(60)));
     }
+
+    @Override
+    public boolean isPawlMotorReady(){
+        return false;
```
aschokking, line 355: ⭐ ⭐ Make sure to mark this as `true` in the unit test contract

src/main/java/competition/subsystems/climber/commands/ReleaseClimberSolenoidCommand.java
```diff
@@ -0,0 +1,48 @@
...
+        System.out.println("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        releaseClimberSolenoid();
+    }
+
+    public void releaseClimberSolenoid(){
+        while (XTimer.getFPGATimestampTime().lt(pawlTimestampEnd)){
```
aschokking, line 39: ⭐ ⭐ ⭐ this needs to live in the subsystem so it's robust. Also can't use `while` loops

aschokking (changes requested):

SeriouslyCommonLib
aschokking, line 1: for posterity, suggest reverting this change so you're not updating SCL when you didn't intend to
`git checkout main -- SeriouslyCommonLib` from the root of your repo while on your branch should do it (and then commit that change)

aschokking (commented):
Apparently the new climber design won't need a pawl anymore so we can probably shelve this PR. Thanks for getting something working so fast so we could test the previous climber (and see that it didn't work)

PR #223, opened 2025-03-02
Add test for Dependency Injection components

stephenjust:
# Why are we doing this?
Students frequently reference the wrong types when doing dependency injection.

# Whats changing?
Add a test that will fail the build if a forbidden class is injected in a way that will impact a real robot.

# How this was tested
- [x] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/test/java/competition/injection/ComponentsTest.java
+++ b/src/test/java/competition/injection/ComponentsTest.java
@@ -1,0 +1,54 @@
+package competition.injection;
+
+import competition.injection.components.DaggerRobotComponent;
+import competition.injection.components.DaggerRobotComponent2023;
+import competition.injection.components.DaggerRobotComponent2024;
+import competition.simulation.MapleSimulator;
+import competition.simulation.MapleSimulator_Factory;
+import org.junit.Test;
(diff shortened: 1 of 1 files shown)


src/test/java/competition/injection/ComponentsTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.junit.Assert.fail;
+
+public class ComponentsTest {
+    @Test
+    public void testRealRobotsDoNotHaveMockDevices() {
+        Set<Type> forbiddenTypes = new HashSet<>();
+        forbiddenTypes.add(MockCANMotorController_Factory.class);
+        forbiddenTypes.add(MockCANMotorController.MockCANMotorControllerFactory.class);
+        forbiddenTypes.add(MapleSimulator_Factory.class);
+        forbiddenTypes.add(MapleSimulator.class);
```
aschokking, line 29: ❤️

aschokking (approved):
Would have saved us a lot of pain this season!

PR #227, opened 2025-03-02
Elevator Safety

kujo27:
# Why are we doing this?

Asana task URL:

# Whats changing?
Added a safety so elevator will not lower when coral arm is greater than 120 degrees.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
+++ b/src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
@@ -3,8 +3,10 @@ package competition.subsystems.elevator.commands;
 import competition.electrical_contract.ElectricalContract;
 import competition.motion.TrapezoidProfileManager;
 import competition.operator_interface.OperatorInterface;
+import competition.subsystems.coral_arm.CoralArmSubsystem;
 import competition.subsystems.elevator.ElevatorSubsystem;
 import edu.wpi.first.units.measure.Distance;
+import edu.wpi.first.wpilibj.Alert;
 import xbot.common.advantage.AKitLogger;
 import xbot.common.command.BaseMaintainerCommand;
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -124,6 +135,12 @@ protected void calibratedMachineControlAction() {
         elevator.setElevatorHeightGoalOnMotor(setpoint);
     }
 
+    private boolean wouldHitBranch(Distance targetValue) {
+        var coralArmAngleDangerous = coralArm.getCurrentValue().gt(Degrees.of(coralArmCollisionAngleDegrees.get()));
+        var elevatorIsMoving = elevator.getTargetValue() != elevator.getCurrentValue();
```
JohnGilb, line 140: This will always be true; you'll have a scenario like:
getTargetValue ==   0.458678758
getCurrentValue == 0.458999212

JohnGilb (changes requested):

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -105,17 +120,21 @@ protected void coastAction() {
     @Override
     protected void calibratedMachineControlAction() {
 
-        var currentValue = elevator.getCurrentValue();
+        var wouldCollide = wouldHitBranch(elevator.getTargetValue());
+        var currentTarget = wouldCollide
+                ? elevator.getCurrentValue()
```
aschokking, line 125: ⭐ ⭐  instead of limiting this to the current value, you could make it where the threshold is so the elevator is allowed to come down until that point?

aschokking (approved):

PR #241, opened 2025-03-05
More careful use of vision suppression

JohnGilb:
# Why are we doing this?
Now that we are using more "Align To April Tag" scenarios, we may end up in a scenario where we are running a bunch of them back-to-back. We still want to pick up vision updates while vision is suppressed. So, when you "suppress vision", this is what will happen:
* Wheels Odometry snaps to Vision Odometry
* Wheels Odometry is now used to report robot pose

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/pose/PoseSubsystem.java
```diff
@@ -111,23 +110,22 @@ protected void updateOdometry() {
...
-            this.aprilTagVisionSubsystem.getAllPoseObservations().forEach(observation -> {
-                fullSwerveOdometry.addVisionMeasurement(
-                        observation.visionRobotPoseMeters(),
-                        observation.timestampSeconds(),
-                        observation.visionMeasurementStdDevs()
-                );
-            });
-        }
+        this.aprilTagVisionSubsystem.getAllPoseObservations().forEach(observation -> {
+            fullSwerveOdometry.addVisionMeasurement(
```
aschokking, line 114: can we rename `fullSwerveOdometry` to something else? the fact that it doesn't have vision in the name makes me always think this is the wheels only version

aschokking: like since the prop is called `useVisionAssistedPose` shouldn't this be `visionAssistedPoseEstimator` or something?

PR #243, opened 2025-03-05
Support the Pigeon 2 IMU on the 2025 robot

stephenjust:
This also tracks accelerations over time, we might be able to use this for collision detection in future.

# Why are we doing this?
The robot has a Pigeon 2 IMU on it. We should at least try to use it!

# Whats changing?
Add support for switching between NavX and Pigeon 2.

The injection here is super ugly, since we default to the NavX in the Common Lib contracts. We'll need some universal factory like the XCANMotorController eventually.

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

stephenjust:
@Team488/core-robot-maintainers Should be safe to review/merge now!

JohnGilb (approved):

aschokking:
Seems like today would be a good time to try adding this?

PR #248, opened 2025-03-05
Loosening interstitial drive commands, setup elastic layout for auto

anthonytrenh:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

JohnGilb:
There's something odd happening here where the drive to face command never actually finishes in sim, and we run into the same "shoving for long durations" problem. I'm taking a look.

PR #252, opened 2025-03-06
Vision pathplanning

Kobeeeef:
# Why are we doing this?

Path planning

Asana task URL:

# Whats changing?

# How this was tested
- [x] tested on robot
- [x] tested in simulator
- [ ] unit tests added

[Screencast from 2025-03-06 23-56-26.webm](https://github.com/user-attachments/assets/fa3b53d0-15b8-47e4-b1ad-cea8ec37b205)
-----

PR feedback legend

 

src/main/java/competition/commandgroups/vision_path/PathToFaceAndScoreCommandGroupFactory.java
```diff
@@ -0,0 +1,60 @@
...
+import edu.wpi.first.wpilibj2.command.ParallelCommandGroup;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import static edu.wpi.first.units.Units.Meters;
+
+public class PathToFaceAndScoreCommandGroupFactory {
+    PathToReefFaceThenAlignCommandGroupFactory
```
aschokking, line 16: 👏🏻 this one is done well with all Providers!

aschokking (changes requested):
Only had time to look at the first ~20%, but here's some initial feedback

src/main/java/competition/commandgroups/vision_path/PathToBestReefBranchLevelThenAlignAndPrepCommandGroupFactory.java
```diff
@@ -0,0 +1,99 @@
...
+                        int cameraIndex = driveToReefFaceCommandUntilAprilTagDetection.curves.getAlignToReefAprilTagOptions().getCamera()
+                                .equals(XTableValues.AprilTagCamera.FRONT_LEFT) ? Cameras.FRONT_LEFT_CAMERA.getIndex() :
+                                Cameras.FRONT_RIGHT_CAMERA.getIndex();
+
+                        alignToReefWithAprilTagCommand.setConfigurations(cameraIndex,
+                                driveToReefFaceCommandUntilAprilTagDetection.curves.getAlignToReefAprilTagOptions()
+                                        .getAprilTagID(),false, -2);
+                        measureDistanceBeforeScoringCommand.setDistanceThreshold(Meters.of(1));
+                    } else {
+                        alignToReefWithAprilTagCommand.end(true); // Not sure if it cancels it or not.
```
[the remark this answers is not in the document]
Kobeeeef: Cancel/End the command in anyway possible, and not make it continue the group?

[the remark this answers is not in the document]
aschokking: If you want to end the whole group, you should call cancel() on the group itself. Instead of returning new group, assign it to a variable first so you can still reference it here and call cancel on it.

SeriouslyCommonLib
aschokking, line 1: ⭐ ⭐ if you didn't actually intend to change SCL, please revert this part of the PR
`git checkout main -- SeriouslyCommonLib`

src/main/java/competition/subsystems/drive/commands/vision_path/DriveVectorSmallCommand.java
```diff
@@ -0,0 +1,97 @@
...
+import edu.wpi.first.math.geometry.Pose2d;
+import xbot.common.command.BaseCommand;
+import xbot.common.controls.sensors.XTimer;
+import xbot.common.math.XYPair;
+
+import javax.inject.Inject;
+
+public class DriveVectorSmallCommand extends BaseCommand {
+    private double start;
+    private double last = 0.25;
```
aschokking, line 14: ⭐ consider calling this `duration` and making it of type `Time` so the units can't be messed up
```suggestion
    private Time duration = Seconds.of(0.25);

File: src/main/java/competition/subsystems/drive/commands/vision_path/DriveVectorSmallCommand.java [hunk 4]
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Whether the command has finished. Once a command finishes, the scheduler will call its end()
+     * method and un-schedule it.
+     *
+     * @return whether the command has finished.
+     */
+    @Override
+    public boolean isFinished() {
+        return (XTimer.getFPGATimestamp() - this.start) >= this.last;
```
aschokking, line 84: ⭐ because this pattern is incredibly common, there's some built in support for it inside wpilib commands already. When you create a command you can add `.withTimeout(blah)` to get a version of that command that will timeout and end then

src/main/java/competition/subsystems/drive/commands/vision_path/DriveVectorSmallCommand.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * The main body of a command. Called repeatedly while the command is scheduled.
+     */
+    @Override
+    public void execute() {
+        if (this.isFinished()) {
+            drive.stop(); // Drive.stop doesnt stop unless called continuously
+            return;
+        }
+        XYPair pair = new XYPair(0.50, 0);
```
aschokking, line 44: ⭐ consider making this 0.5 amount a `DoubleProperty` so it's easier to tune

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToLocationCommand.java
```diff
@@ -0,0 +1,134 @@
...
+
+public class PathDriveToLocationCommand extends SwerveBezierTrajectoryBase {
+    Pose2d target;
+
+    private final VisionCoprocessorCommander commander;
+
+    private static boolean useBackupPointToPoint = false;
+
+    private XTableValues.TraversalOptions traversalOptions;
+    private int safeInches = 10;
```
aschokking, line 31: ⭐ Consider using a `Distance` value so the units are baked in. Also generally wouldn't suggest `int` for a value like this that you could want to have a decimal component to

```suggestion
    private Distance safeDistance = Inches.of(10);

File: src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToLocationCommand.java [hunk 7]
```diff
@@ -0,0 +1,134 @@
...
+            message.setArguments(additionalArguments);
+        }
+        curves = commander.requestBezierPathWithOptions(
+                message
+                        .build(),
+                3000, TimeUnit.MILLISECONDS); // When should it give up and return
+        // null for any reason?
+
+        if (curves == null) {
+            useBackupPointToPoint = true;
```
aschokking, line 109: ⭐ I like the idea of this a lot, but maybe this should live somewhere more central, like on the CoprocessorCommunicationSubsystem? So other commands can take advantage of it too and it doesn't need to be static (which makes unit testing more awkward)

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToNearestCoralStationSectionCommandCommand.java
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Target position for the coral station section.
+     */
+    Pose2d targetCoralStationSection;
+    Pose2d destinationPose;
+
+    /**
+     * Distance from the center of the robot to the outer bumper, in meters.
+     */
+    double radiusOfRobot;
```
aschokking, line 45: ⭐ if you used a `Distance` here the units would be handled for you

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToNearestCoralStationSectionCommandCommand.java
```diff
@@ -0,0 +1,161 @@
...
+        this.driveSubsystem = drive;
+        this.electricalContract = electricalContract;
+    }
+
+    /**
+     * Determines the nearest coral station pose based on detected AprilTags.
+     *
+     * @return The estimated pose of the coral station.
+     */
+    private Pose2d getCoralStationPose() {
```
aschokking, line 98: ⭐ ⭐ I'm pretty sure there's already code to do this, might see about re-using what's there? `pose.getClosestCoralStation()`

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToNearestCoralStationSectionCommandCommand.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Initializes the command, setting the target destination based on the
+     * nearest coral station section.
+     */
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+        targetCoralStationSection = this.getCoralStationPose();
+        var deltaTranslation = new Translation2d(
```
aschokking, line 131: ⭐ consider logging the coral station it picked, maybe moving the init log line here and then you can include the station in the init line?

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToNearestCoralStationSectionCommandCommand.java
```diff
@@ -0,0 +1,161 @@
...
+                        .setFinalRotationDegrees(destinationPose.getRotation().getDegrees())
+                        .build());
+        super.initialize();
+    }
+
+    @Override
+    public boolean isFinished() {
+        aKitLog.record("distanceToGoal", pose.getCurrentPose2d().getTranslation().getDistance(destinationPose.getTranslation()));
+        return super.isFinished() 
+        || (destinationPose != null && pose.getCurrentPose2d().getTranslation().getDistance(destinationPose.getTranslation()) < 0.1016);
```
aschokking, line 155: ⭐ suggest pulling magic numbers like this out into named variables, eg `final static Distance goalThreshold = Meters.of(0.1016)`

aschokking (approved):
Thanks for tackling all the previous feedback 🎉 I'm really excited to see this go in soon

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToNearestCoralStationSectionCommandCommand.java
Kobeeeef: John (student) added this method recently. I've had this one for a long time and have tested it multiple times. His uses Landmarks which if you already know are just points that were retrieved from driving around in simulator. This one uses the layout and electrical contract robot sizing to get a real actual math accurate pose rather then just a pose from simulator then flipped around over.

SeriouslyCommonLib
Kobeeeef: All I did was update from SCL main? I am already on the main branch.

aschokking: That's the problem, you're updating the version of SCL that this repo is pointing at to SCL's `main`, which is changing what main of this repo is pointing at. We shouldn't be changing that unintentionally, there could be a breaking change in SCL main that requires an update in this repo to be done before it's SCL version is updated.

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToNearestCoralStationSectionCommandCommand.java
aschokking: I welcome your pitch for improving the current way, but in the short-run we should avoid having multiple ways of doing the same thing in the codebase. If we get to the event and end up noticing things are a bit off from the theoretical model (be that in the sim or from idealized math) we'll now have to fix it in both locations.

SeriouslyCommonLib
Kobeeeef: Ahh, my apologies. I assumed we were always using the latest as I thought that was what would be working. I'll keep that in mind.

PR #253, opened 2025-03-06
Vision/creeper

Kobeeeef:
# Why are we doing this?

Align to reef creeper command.

This is a draft, please review first. I will add sim video when i can.

Asana task URL:

# Whats changing?

Adding align with creeper command.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
@@ -141,10 +141,10 @@ public class AlignWithCreeperCommand extends BaseCommand {
             Integer rightDistance = this.rightDistanceSubscriber.getAsInteger(null);
 
             if (isCenteredConfidently == null) {
-                // There is no new data
+                // No new vision data received; do nothing.
                 return;
             } else if(isCenteredConfidently) {
-                // The robot is already aligned.
+                // Alignment is achieved; stop any drive movement.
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,268 @@
...
+            // Clamp the drive power to the safe range of [-1, 1].
+            drivePower = Math.max(-1.0, Math.min(1.0, drivePower));
+
+            // Reverse the drive power sign if misalignment is to the left to ensure
+            // proper movement direction.
+            if (offToTheLeft) {
+                drivePower = -drivePower;
+            }
+            // Create the drive command vector (Only drive power along the Y-axis,
+            // side-to-side).
```
Ademfcan, line 209: will this move on the y axis relative to the tag? The robots orientation might not be aligned parallel to the tag

Kobeeeef: when you're that close to the reef, you have no accurate pose bc no April tag. You only want to move robot relative left/right. The align to april tag is most of the time parallel to the April tag, otherwise our creeper would already have no use as it would see straight noise data from other lines on the field.

PR #257, opened 2025-03-07
Instrument commands/command groups with Alerts for running

aschokking:
# Why are we doing this?

To see when they start/stop.  Depends on a SCL PR.

Asana task URL:

# Whats changing?

- Change our autonomous commands with groups to extend the Base* versions instead
- Make sure relevant commands call super for init/end so Alert gets updated

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

image

-----

PR feedback legend

 

aschokking:
Stephen has a better approach to this using the Scheduler

PR #259, opened 2025-03-07
align with creeper command

Ademfcan:
# Why are we doing this?
The align to creeper command allows us to build a layer of accuracy on top of the regular align to april tag command by also allowing us to align without directly detecting the tag. This is important because when we are in the final stages of alignment the camera is too close to the april tag to try to detect the whole tag. But we can instead align using the black bars of the tag, to ensure scoring
Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,351 @@
...
+            log.info("FORCE STOP");
+            drive.stop();
+            return;
+        }
+        
+
+        
+        // hack to add some periodiciy to the command
+        runIter++;
+        if(runIter < ITERSPERRUN.get()){
```
aschokking: ❓ What's this doing? I'm kinda confused

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,351 @@
...
+        try {
+            this.isCenteredConfidently = this.isCenteredConfidentlySubscriber.getAsBoolean(null);
+            Integer leftDistance = this.leftOffsetPixelsSubscriber.getAsInteger(null);
+            Integer rightDistance = this.rightOffsetPixelsSubscriber.getAsInteger(null);
+            log.info(String.format("Loff %d Roff %d ifConf %b" , leftDistance,rightDistance,isCenteredConfidently));
+            
+            if (isCenteredConfidently == null) {
+                // No new vision data received; do nothing.
+                log.warn("Centered confidently is returning null!");
+                FORCESTOP = true;
```
aschokking: ⭐ ⭐  if you're cancelling, I think forcestop won't do anything because this command won't get to run again after this return?

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,351 @@
...
+            // add sign (right is negative)            
+            if (offToTheLeft) {
+                error = -error;
+            }
+
+            
+
+            // Scale the normalized error using the drive gain to compute the drive
+            // power.
+            double drivePower = driveGain.get() * error;
```
aschokking, line 258: ⭐ ⭐ ⭐ We'd suggest using our PIDManager here so you get P I and D and properties to easily tune them

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,351 @@
...
+
+            
+            
+
+            // sim issue where if drive power is below a certain amount, the robot does not move.
+            // if(Math.abs(drivePower) < 0.02){
+            //     drivePower = Math.copySign(0.02, drivePower);
+            // }
+
+            log.info("Driving with power: " + drivePower);
```
aschokking, line 271: ⭐ ⭐ ⭐ Please remove this logging, it'll spam the logs too much. the akit logger can be a better place to put this kind of number, you can then graph it over time

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,351 @@
...
+            
+            // Create the drive command vector (Only drive power along the Y-axis,
+            // side-to-side).
+            XYPair pair = new XYPair(0, drivePower);
+            this.drive.move(pair, 0);
+
+            // should we always stop the drive after a move command?
+            // this.drive.stop();
```
aschokking, line 279: ⭐ ⭐ the drive will do the last thing it was requested to do in a loop, so if you call stop after move stop will win

aschokking (changes requested):

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+        if(runIter < itersPerRUn.get()){
+            return;
+        }
+        runIter = 0; // reset
+
+        try {
+            this.isCenteredConfidently = this.isCenteredConfidentlySubscriber.getAsBoolean(null);
+            Integer leftDistance = this.leftOffsetPixelsSubscriber.getAsInteger(null);
+            Integer rightDistance = this.rightOffsetPixelsSubscriber.getAsInteger(null);
+            log.info(String.format("Loff %d Roff %d ifConf %b" , leftDistance,rightDistance,isCenteredConfidently));
```
JohnGilb, line 197: logging every loop is a perf hit

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+
+        
+        // hack to add some periodiciy to the command
+        runIter++;
+        if(runIter < itersPerRUn.get()){
+            return;
+        }
+        runIter = 0; // reset
+
+        try {
```
JohnGilb, line 193: try/catch should be avoided especially in commonly-run parts like `execute()`, as exceptions are very expensive on the machine.

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+            log.info("FORCE STOP");
+            drive.stop();
+            return;
+        }
+        
+
+        
+        // hack to add some periodiciy to the command
+        runIter++;
+        if(runIter < itersPerRUn.get()){
```
JohnGilb, line 188: Why?

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+        }
+        runIter = 0; // reset
+
+        try {
+            this.isCenteredConfidently = this.isCenteredConfidentlySubscriber.getAsBoolean(null);
+            Integer leftDistance = this.leftOffsetPixelsSubscriber.getAsInteger(null);
+            Integer rightDistance = this.rightOffsetPixelsSubscriber.getAsInteger(null);
+            log.info(String.format("Loff %d Roff %d ifConf %b" , leftDistance,rightDistance,isCenteredConfidently));
+            
+            if (isCenteredConfidently == null) {
```
JohnGilb, line 199: This seems like it leaves us very vulnerable to a temporary (e.g. 1-frame) loss of data.

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+    /**
+     * Checks if the alignment is completed.
+     *
+     * @return true if the vision system indicates the robot is confidently
+     * aligned, false otherwise.
+     */
+    @Override
+    public boolean isFinished() {
+        // us being centered is only valuable if we arent currently moving
+        return this.isDriveStopped() && this.isCenteredConfidently;
```
JohnGilb, line 310: if you add `|| forceStop` here then you can avoid all those `cancel()` calls

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+            } else if (rightDistance == -1) {
+                offToTheLeft = true; // we dont see right  edge, means left side is "too" visible eg off to the left
+            } else {
+                offToTheLeft = leftDistance < rightDistance; // we see both edges, so the one that is "closer to the center" is the one we are off by
+            }
+            // Choose the pixel error value from the side indicating misalignment.
+            
+            // Calculate the error as a function of the left and right distance from the center.
+            double error = costFunc(leftDistance, rightDistance);
+            log.info("Normalized error: " + error);
```
JohnGilb, line 237: same comment about aggressive logging in `execute()`

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,352 @@
...
+            }
+
+            
+
+            // Scale the normalized error using the drive gain to compute the drive
+            // power.
+            double drivePower = driveGain.get() * error;
+
+            // Clamp the drive power to the safe range of [-1, 1].
+            drivePower = Math.max(-1, Math.min(1, drivePower));
```
JohnGilb, line 261: A power of `1` will move the robot an enormous amount. This should likely be smaller by default, and a Property. (I think Alex told you about the PIDPropertyManager)

Ademfcan:
> One more note - I believe simulating this requires extra setup on computers. @Ademfcan , can you share the instructions needed to verify this behavior in sim?


To test this, you will need the Team488/advantagescope-mjpeg and Team488/Alt repos.  Additionally you will need to run an xtables server jar from Kobeeeef/XTABLES releases. In the advantagescope-mjpeg repository's releases there is a .exe file that should be all you need to run. On the Alt side of things you will need python3.9 and installing the minimalalignmentenv.txt requirements (there is currenty a precommit issue where there is a null terminator in the end of .txt files, so you might have to remove one from this) . Then from inside of the src directory, you want to run alignmentcheck.py, that should already be configured to read from the simulation endpoint for the front right at camera.  Then in the robot code if you check out this branch and simulate, i belive currently the x button is mapped to align with the right camera. I added the y button too, but on the python side no data will be coming in for that camera. That should all work, but i have been having issues with pip enviornments not properly working together so you might have to tweak versions. Once i have some time i will finish the sim docs 🤞

Kobeeeef:
> One more note - I believe simulating this requires extra setup on computers. @Ademfcan , can you share the instructions needed to verify this behavior in sim?

https://team488.gitbook.io/

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
```diff
@@ -0,0 +1,289 @@
...
+        // log is negative when the input is less that 1,
+        // to make this never happen, add 1
+        double errFunc = Math.log(err/logErrScalar.get()+1)/logScalar.get(); 
+        return Math.max(-maxError.get(), Math.min(errFunc, maxError.get()));
+    }
+
+    @Override
+    public boolean isFinished() {
+        // us being centered is only valuable if we arent currently moving
+        return this.isDriveStopped() && (this.isCenteredConfidently == null || this.isCenteredConfidently);
```
aschokking, line 261: ⭐ it might help with debugging if you logged here why it was ending (ie, was it confidently centered or was that null and it gave up?)

src/main/java/competition/subsystems/drive/logic/AlignWithCreeperCalculator.java
```diff
@@ -0,0 +1,282 @@
...
+    private final AKitLogger aKitLog;
+    private final PIDManager pidManager;
+
+    final DriveSubsystem drive;
+    final CoprocessorCommunicationSubsystem coprocessorCommunicationSubsystem;
+
+    private final int tunedWidth = 960; // what resolution the pid tuning was done at
+    private final int tunedHeight = 720; // what resolution the pid tuning was done at
+
+    private final String tableLeftDistance = "verticalEdgeLeftDistancePx";
```
aschokking, line 35: ⭐ all these const values are duplicated with your other command, consider maybe making them public so you can reference them from here instead?

src/main/java/competition/subsystems/drive/logic/AlignWithCreeperCalculator.java
```diff
@@ -0,0 +1,282 @@
...
+                error = -error;
+            }
+
+
+        }
+
+        aKitLog.record("creeper error from cost function: ",error);
+
+        // power.
+        double drivePower = driveGain.get() * pidManager.calculate(0, error);
```
aschokking, line 227: ⭐ I'm a little confused about why you need driveGain in addition to the PID?

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommandWCalc.java
```diff
@@ -0,0 +1,79 @@
...
+import xbot.common.math.XYPair;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.PropertyFactory;
+import xbot.common.properties.StringProperty;
+import xbot.common.subsystems.drive.control_logic.HeadingModule;
+import competition.subsystems.drive.logic.AlignWithCreeperCalculator;
+import javax.inject.Inject;
+
+
+public class AlignWithCreeperCommandWCalc extends BaseCommand {
```
aschokking, line 27: ⭐ ⭐ Would recommend Command being the suffix here to match our other commands
```suggestion
public class AlignWithCreeperWCalcCommand extends BaseCommand {

File: src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java [hunk 17]
```diff
@@ -0,0 +1,289 @@
...
+    private String hostname;
+    private Cameras camera;
+    // set defaults for now
+    private int currentCamHres = this.tunedWidth;
+    private int currentCamVres = this.tunedHeight;
+
+    protected Boolean isCenteredConfidently = false;
+
+    @Inject
+    public AlignWithCreeperCommand(AprilTagVisionSubsystemExtended vision,
```
aschokking, line 77: ❓ why does this exist in addition to the calc version?

aschokking (changes requested):
Overall looking good, excited to get this tested! 

The main blocking thing is changing the drivers gamepad

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommand.java
Ademfcan: Will remove, wanted to persist the original till i finished the calculator version

src/main/java/competition/subsystems/drive/commands/AlignWithCreeperCommandWCalc.java
Ademfcan: 👍

aschokking (approved):
You've got some merge conflicts from main to handle, but otherwise changes look good. Just make sure SCL doesn't look like it's being changed in this PR after you merge main

PR #266, opened 2025-03-07
Hardcoded bezier for auto

Rongrrz:
# Why are we doing this?
A smoother curve movement

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

https://github.com/user-attachments/assets/178ca549-1f06-48d6-b6c0-5f9d2e582d1b

-----

PR feedback legend

 

JohnGilb:
I like this - we should revisit this on Tuesday post-competition since we won't have a ton of time to try out new routines at the event.

aschokking:
@kujo27 is this something that should be our fallback if the curves aren't working?

PR #267, opened 2025-03-07
Feature/more reef alignment points

JohnGilb:
# Why are we doing this?
Allows us to align from closer (at the cost of moving a bit slower)
Asana task URL:

# Whats changing?

# How this was tested
- [x] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

aschokking (approved):
Makes sense!

PR #271, opened 2025-03-08
Hermite Splines

JohnGilb:
# Why are we doing this?

Was thinking about a different kind of spline follower, pathing along the spline directly rather than creating several intermediate points and re-using the SimpleSwerveTrajectory. Overall felt a little easier to tune.

Is also about 80% of the way to hooking up the NeoTrellis buttons to scoring/collection command groups using said splines, which should allow for a relatively simple drop-in of a more advanced source of splines (e.g. the Orin)

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/motion/CubicHermiteSpline.java
+++ b/src/main/java/competition/motion/CubicHermiteSpline.java
@@ -1,0 +1,312 @@
+package competition.motion;
+
+    import edu.wpi.first.math.geometry.Translation2d;
+
+    /**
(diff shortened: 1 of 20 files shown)


src/main/java/competition/motion/CubicHermiteSpline.java
aschokking, line 1: Is there an off the shelf library we could use for this math?

PR #273, opened 2025-03-09
Feature/teleop no retries and rumble when done

JohnGilb:
# Why are we doing this?
Driver Request
Asana task URL:

# Whats changing?
* Teleop align to reef will never retry
* When align to reef successfully finishes, it rumbles both gamepads.

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -396,6 +394,8 @@ public AlignCameraToAprilTagAdvice getXYPowersAlignToAprilTag(Pose2d currentPose
                 // If we've been shoving for a while, we're done.
                 if (XTimer.getFPGATimestamp() - shoveStartTime > shoveDuration.get()) {
                     activity = Activity.Complete;
+                    oi.operatorGamepad.getRumbleManager().rumbleGamepad(1,.75);
```
aschokking, line 397: ⭐ this might be an aggressive rumble for something that might be running for multiple seconds as they score? Also maybe the duration should be shorter so when they let go of the button it stops sooner?

PR #293, opened 2025-03-15
Changing the coral arm angle in sim

Glowdisk:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
+++ b/src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
@@ -15,9 +15,9 @@ public class CoralArmSimConstants {
     public static final Distance armLength = Meters.of(0.5);
     // the frame of reference for these angles is 0' right, 90' up, 180' left, 270' down
     // so we have 225 as the starting angle which is 0' in arm relative terms
-    public static final Angle minAngleRads = Degrees.of(225 - 180);
+    public static final Angle minAngleRads = Degrees.of(237 - 180);
     public static final Angle armEncoderAnglePerRotation = Degrees.of(6.94444);
     public static final Angle angleAtRobotZero = Degrees.of(225);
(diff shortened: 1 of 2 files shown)


src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
```diff
@@ -15,9 +15,9 @@ public class CoralArmSimConstants {
     public static final Distance armLength = Meters.of(0.5);
     // the frame of reference for these angles is 0' right, 90' up, 180' left, 270' down
     // so we have 225 as the starting angle which is 0' in arm relative terms
-    public static final Angle minAngleRads = Degrees.of(225 - 180);
+    public static final Angle minAngleRads = Degrees.of(237 - 180);
     public static final Angle armEncoderAnglePerRotation = Degrees.of(6.94444);
     public static final Angle angleAtRobotZero = Degrees.of(225);
```
aschokking, line 20: ⭐ ⭐ I think we need to update this one as well to be 237 so the encoder values are faked correctly?

aschokking (approved):

src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
Glowdisk: should it be 225?  this what i got in the sim when i plug in 225.

image

aschokking: I guess it's okay then, thanks for trying

aschokking:
It turned out there were a few more spots that needed to be updated to make this work right end to end, I created another PR that has this plus the extra stuff in it: https://github.com/Team488/TeamXbot2025/pull/313

Thanks for helping tackle the first part David!

PR #295, opened 2025-03-15
Created camera light states except for creeper

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -121,12 +135,25 @@ public LightsStateMessage getCurrentState() {
             if (!Objects.equals(autonomousCommandSelector.getProgramName(), "EmergencyAutonomousCommand")) {
                 currentState = LightsStateMessage.RobotDisabledAuto;
             }
+            if (!visionSubsystem.areAllCamerasConnected()) {
```
aschokking, line 138: ⭐ ⭐  this should probably be higher in the disabled list. Otherwise if they set a custom auto we won't know the lights are broken and that maybe they should just run the default after all?

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -121,12 +135,25 @@ public LightsStateMessage getCurrentState() {
...
-            if (coralScorerSubsystem.confidentlyHasCoral() && coralArmSubsystem.getIsTargetAngleScoring()
-                    && coralArmSubsystem.isMaintainerAtGoal() && elevatorSubsystem.isMaintainerAtGoal()) {
-                currentState = LightsStateMessage.ReadyToScore;
+            if (activity == AlignCameraToAprilTagCalculator.Activity.TerminalApproach) {
+                if (!visionSubsystem.isCameraConnected(targetCameraID)) {
+                    currentState = LightsStateMessage.TargetCameraUnavailable;
+                } else {
+                    currentState = LightsStateMessage.CurrentlyAligning;
+                }
+            } else if (activity == AlignCameraToAprilTagCalculator.Activity.Complete) {
```
aschokking, line 148: ⭐ ⭐ ⭐ I think it's going to be stuck in complete for the whole match unless the driver starts another alignment action because nothing ever clears complete? Maybe after scoring?

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -121,12 +135,25 @@ public LightsStateMessage getCurrentState() {
...
+                currentState = LightsStateMessage.FinishedAligning;
             } else if (coralScorerSubsystem.confidentlyHasCoral()) {
                 currentState = LightsStateMessage.CoralPresent;
+            } else if (coralScorerSubsystem.confidentlyHasCoral() && coralArmSubsystem.getIsTargetAngleScoring()
```
aschokking, line 152: ⭐ ⭐ I _think_ this won't get a chance to go if activity == complete is true above right which would be true when we're attempting to score? Did you try this out in the simulator? I could be wrong..

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -143,6 +170,13 @@ public LightsStateMessage getState() {
         return state;
     }
 
+    public void updateActivity(AlignCameraToAprilTagCalculator.Activity activity) {
+        this.activity = activity;
+    }
+
+    public void updateTargetCameraID(int targetCameraID) {
```
aschokking, line 177: ⭐  Since you want this information at the same time as activity getting updated, why not combine them into 1 method? That way nobody can forget to provide both at the same time, it's enforced:

```suggestion
public void updateActivity(AlignCameraToAprilTagCalculator.Activity activity, int targetCameraID) {

File: src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java [hunk 5]
```diff
@@ -94,4 +96,13 @@ public int getTargetAprilTagID(Landmarks.ReefFace reefFace) {
         Pose2d targetReefFacePose = PoseSubsystem.convertBlueToRedIfNeeded(Landmarks.getReefFacePose(reefFace));
         return getTargetAprilTagID(targetReefFacePose);
     }
+
+    public boolean areAllCamerasConnected() {
```
aschokking, line 100: 👏🏻

aschokking (changes requested):
Please do some testing in the sim to make sure the output light state matches what you expect as you try to drive + score with the robot (bring a friend if you need 2 drivers with 2 gamepads)

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -42,7 +50,14 @@ public enum LightsStateMessage{
         CoralPresent(4),
         RequestCoralFromHuman(5),
         Victory(6), 
-        ReadyToScore(7);
+        //ReadyToScore(7),
```
aschokking, line 53: ⭐ why don't we have ready to score anymore?

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -118,17 +135,37 @@ public LightsStateMessage getCurrentState() {
             if (DriverStation.getMatchTime() > 230 && DriverStation.isTeleop()) {
                 currentState = LightsStateMessage.Victory;
             }
+            if (!visionSubsystem.areAllCamerasConnected()) {
+                currentState = LightsStateMessage.DisabledCameraUnavailable;
+            }
             if (!Objects.equals(autonomousCommandSelector.getProgramName(), "EmergencyAutonomousCommand")) {
```
aschokking, line 141: ⭐ ⭐ this needs to be an `else` if we want the cameras not working thing to dominate?

src/main/java/competition/subsystems/lights/LightSubsystem.java
```diff
@@ -118,17 +135,37 @@ public LightsStateMessage getCurrentState() {
...
+                    currentState = LightsStateMessage.TargetCameraUnavailable;
+                } else {
+                    currentState = LightsStateMessage.CurrentlyAligning;
+                    recentlyAligned = true;
+                }
+            } else if (activity == AlignCameraToAprilTagCalculator.Activity.Complete && recentlyAligned) {
+                currentState = LightsStateMessage.FinishedAligning;
+
+                if (lastStateTime == -1) {
+                    lastStateTime = XTimer.getFPGATimestamp();
```
aschokking, line 156: ❓ what's the idea behind last state time and why do we only update it here?

aschokking (approved):

PR #304, opened 2025-03-16
Only consume the freshest LaserCAN updates

stephenjust:
# Why are we doing this?
LaserCAN updates happen less frequently than the robot code loop. We don't want to calibrate positions off of stale data, so make sure that we only report LaserCAN measurements while they're reasonably fresh.

# Whats changing?

# How this was tested
- [x] tested on robot
- [x] tested in simulator
- [ ] unit tests added

image

-----

PR feedback legend

 

aschokking (approved):
Worth trying on the robot! It's complicated (because it's a complicated problem) so there's lots of space for something to not work

PR #313, opened 2025-03-20
Update coral arm in sim and superstructure vis

aschokking:
# Why are we doing this?

Asana task URL:

# Whats changing?

- Disable the arm absolute encoder in unit test contract, we don't have the device and probably never will
- Update the superstructure vis so that 0' arm angle is now the new bottom spot
- Change the params of the coral arm sim to match the new positioning

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

##  Video/screenshots (from simulator or live robot)

https://github.com/user-attachments/assets/b7b78b84-67e6-4b73-9b22-a8e58b69bdc4

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
+++ b/src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
@@ -10,14 +10,16 @@ import edu.wpi.first.units.measure.Mass;
 
 public class CoralArmSimConstants {
     // TODO: this is all from the wpilib example but doesn't reflect our real math
-    public static final double armReduction = 200;
+    public static final double armReduction = 50;
     public static final Mass armMass = Kilogram.of(8.0); // Kilograms
     public static final Distance armLength = Meters.of(0.5);
+
     // the frame of reference for these angles is 0' right, 90' up, 180' left, 270' down
     // so we have 225 as the starting angle which is 0' in arm relative terms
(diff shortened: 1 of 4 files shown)


src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
```diff
@@ -10,14 +10,16 @@
 
 public class CoralArmSimConstants {
     // TODO: this is all from the wpilib example but doesn't reflect our real math
-    public static final double armReduction = 200;
+    public static final double armReduction = 50;
```
aschokking, line 13: This change makes the arm behave a little bit closer to the real one, the previous value made it so the arm barely budged when unpowered

src/main/java/competition/simulation/coral_arm/CoralArmSimConstants.java
```diff
@@ -10,14 +10,16 @@
...
     public static final Mass armMass = Kilogram.of(8.0); // Kilograms
     public static final Distance armLength = Meters.of(0.5);
+
     // the frame of reference for these angles is 0' right, 90' up, 180' left, 270' down
     // so we have 225 as the starting angle which is 0' in arm relative terms
-    public static final Angle minAngleRads = Degrees.of(225 - 180);
```
aschokking, line 18: cleaned up this code a bit so it doesn't repeat values

PR #314, opened 2025-03-20
Motion magic on elevator

junyu101:
# Why are we doing this?
Potentially faster elevator movement, I have no clue on how accurate the simulator reflects the speed of the elevator in real life, but the max acceleration and jerk values are properties so its tunable.
Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x ] tested in simulator
- [ ] unit tests added

https://github.com/user-attachments/assets/6ad9fb4a-c27e-4fbd-81ba-e55494a8b9cd

-----

PR feedback legend

 

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -379,6 +389,19 @@ public void trimElevatorDown(){
...
 
+    public void toggleMotionMagic(){
+        motionMagicEnabled = !motionMagicEnabled;
+    }
+
+    public boolean isMotionMagicEnabled(){
+        return motionMagicEnabled;
+    }
+
+    public void configureMotionMagicConstraints(){
```
JohnGilb, line 400: :star: I recommend looking at how we tune the P/I/D values for the motors, where property changes are picked up at runtime rather than having to reboot the robot with your current approach. It will make it much easier to tune the elevator.

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -348,7 +356,9 @@ public void setElevatorDeltaFromCurrentHeight(Distance heightDelta) {
         var deltaRotations = Rotations.of(heightDelta.in(Meters) * rotationsPerMeter.get());
         masterMotor.setPositionTarget(
                 masterMotor.getPosition().plus(deltaRotations),
-                XCANMotorController.MotorPidMode.Voltage);
+                motionMagicEnabled
+                ? XCANMotorController.MotorPidMode.TrapezoidalVoltage
```
aschokking, line 360: ⭐ ⭐ We might want to try and S-curve rather than Trapezoid to have smoother starts/ends? https://v6.docs.ctr-electronics.com/en/latest/docs/api-reference/device-specific/talonfx/motion-magic.html

aschokking (approved):
Approving, but let's not actually merge this into `main` until after this competition just in case

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
aschokking: Ah apparently this is what we've mapped trapezoid voltage to, nvm

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -61,7 +62,10 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
     public final DoubleProperty powerNearUpperLimitThreshold;
     public final DoubleProperty powerWhenBottomSensorHit;
 
+    public final DoubleProperty motionMagicAcceleration;
+    public final DoubleProperty motionMagicJerk;
 
+    public boolean motionMagicEnabled;
```
stephenjust, line 68: ⭐⭐Make this a boolean property so we can set it from Elastic

stephenjust (approved):

stephenjust:
> Approving, but let's not actually merge this into `main` until after this competition just in case

I would disagree on that. Since it's toggle-able, I think it's pretty safe to merge.

aschokking:
> > Approving, but let's not actually merge this into `main` until after this competition just in case
> 
> I would disagree on that. Since it's toggle-able, I think it's pretty safe to merge.

Balanced against the risk there was a bug with the toggleability and it actually regresses something for our first match 🙏🏻

PR #330, opened 2025-03-28
RIP ORACLE

Glowdisk:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

aschokking (approved):
Looks good, this will remove a bunch of properties and we can always revert this PR if we decide we do want to pursue the oracle for a later event at worlds

PR #331, opened 2025-03-28
Added gravity power amount to manual control

AlanYuan99:
# Why are we doing this?

Asana task URL:

# Whats changing?

What is the proper way to add my new property to the humanControlAction() override?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -97,7 +100,7 @@ protected void initializeMachineControlAction() {
 
     @Override
     protected void coastAction() {
-        elevator.setPower(0);
+        elevator.setPower(manualGravityPower.get());
```
aschokking, line 103: ⭐ worth adding a comment here about why we're doing this, ie during coast we don't want the elevator to drop dramatically like it would with 0 power

src/main/java/competition/subsystems/elevator/commands/ElevatorMaintainerCommand.java
```diff
@@ -150,7 +153,7 @@ private void attemptCalibration(){
 
     @Override
     protected void humanControlAction() {
-        super.humanControlAction();
+        super.humanControlAction() + manualGravityPower.get();
```
aschokking, line 156: ⭐ ⭐ ⭐ this isn't going to work, and actually won't compile.

` super.humanControlAction()` calls the parent class's implementation, which is going to send the current joystick value as power to the motor. Instead of calling that code we want to replace it here with your own logic that does what we want. 

What we want is to set to the motor the power of getHumanInput() + manualGravityPower.get()

aschokking (dismissed):

stephenjust (approved):

PR #332, opened 2025-03-28
Silence loop overrun logs!

stephenjust:
# Why are we doing this?
The full loop overrun print is expensive. Trim it down!

# Whats changing?

# How this was tested
- [x] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

image

-----

PR feedback legend

 

src/main/java/competition/aspects/DriverStationOverrides.java
```diff
@@ -0,0 +1,23 @@
...
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+
+@Aspect
+public class DriverStationOverrides {
+    public static final Logger log = LogManager.getLogger(DriverStation.class);
+
+    @Around("execution(* edu.wpi.first.wpilibj.DriverStation.reportWarning(java.lang.String, boolean))")
+    public void reportWarningOverride(ProceedingJoinPoint joinPoint) {
+        log.warn(joinPoint.getArgs()[0]);
```
aschokking, line 16: ⭐ I'm just wondering if we maybe want to be able to toggle this so when we're doing debugging we can see all the epoch info? Vs when we're in a match on FMS we definitely don't want the prod robot code slowed down

aschokking (approved):

PR #333, opened 2025-03-29
tune -11 inches from robot bumper

Kobeeeef:
# Why are we doing this?
Screenshot from 2025-03-28 20-00-16

Asana task URL:

# Whats changing?
tune values -11 inches into coral station, this way it starts slowing down perfectly into the coral station then backs in as another precaution

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

aschokking (approved):
Let's maybe try this branch out on the robot before we merge it into main?

PR #335, opened 2025-03-29
Branch offsets

kujo27:
# Why are we doing this?
Allows us to have specific horizontal branch offsets
Asana task URL:

# Whats changing?
auto elastic layout, OperatorCommandMap, AlignCameraToAprilTagCalculator

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
+++ b/src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
@@ -99,6 +99,7 @@ public class AlignCameraToAprilTagCalculator {
 
     final DoubleProperty globalHorizontalOffsetInches;
     private double globalTemporaryHorizontalOffsetInches;
+    private double branchOffsetMeters;
 
     double lastKnownHorizontalErrorMeters = 999999;
     double shoveStartTime = 0;
@@ -226,11 +227,7 @@ public class AlignCameraToAprilTagCalculator {
                 isCameraBackwards
         );
 
-
-        double branchOffsetMeters = 0;
         branchOffsetMeters = (targetCameraID == 0 ? branchBOffsetHashMap : branchAOffsetHashMap).get(targetAprilTagID).get();
-
-        akitLog.record("lastBranchOffset-M", branchOffsetMeters);
 
         this.alignmentPointOffset = new Translation2d(
                 alignmentPointOffset.getX(),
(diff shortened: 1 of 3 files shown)


src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -113,6 +116,9 @@ public enum ApproachMode {
     private Translation2d coralStationPreShovePoint;
     boolean retryActive;
 
+    Map<Integer, DoubleProperty> branchAOffsetHashMap = new HashMap<>();
```
aschokking, line 119: ⭐ ⭐ Instead of two maps, you could have the branch be part of the key. You could have a `record` that combines an april tag id with a landmark branch

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -511,4 +521,31 @@ private Translation2d getCoralStationPreShovePoint() {
...
+
+    private void initializeBranchOffsets(PropertyFactory pf) {
+        int[] tagIds = {6, 7, 8, 9, 10, 11, 17, 18, 19, 20, 21, 22};
+        for (int tagId : tagIds) {
+            branchAOffsetHashMap.put(tagId, pf.createPersistentProperty("BranchAOffset-Tag" + tagId, 0.0));
+            branchBOffsetHashMap.put(tagId, pf.createPersistentProperty("BranchBOffset-Tag" + tagId, 0.0));
+        }
+    }
+
+    private void setBranchOffsetMeters(int targetCameraID) {
```
aschokking, line 533: ❓ Can you help me understand how this would work in practice? My understanding is that when the robot is flush (and we can measure the offset) with the reef we can't see any april tags

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
```diff
@@ -473,7 +483,7 @@ private void updateFinalTargetState(Pose2d currentPose) {
         Optional<Translation2d> aprilTagData = aprilTagVisionSubsystem.getRobotRelativeLocationOfAprilTag(targetCameraID, targetAprilTagID);
 
         if (aprilTagData.isPresent()) {
-            lastKnownHorizontalErrorMeters = aprilTagData.get().getY() + getHorizontalTrimAdjustmentMeters();
+            lastKnownHorizontalErrorMeters = aprilTagData.get().getY() + getHorizontalTrimAdjustmentMeters() + branchOffsetMeters;
```
aschokking, line 486: ⭐ ⭐ is this really the only place we need to update? I thought there was at least 1 other spot getHorizontalTrimAdjustmentMeters() was being called

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
kujo27: I made this as a way to measure the offsets quickly with a press of a button, but this requires the robot to see the tag  (a couple inches away from the tag).

src/main/java/competition/subsystems/drive/logic/AlignCameraToAprilTagCalculator.java
kujo27: It was being called when creating the alignmentPointOffset (Translation2d) but I have updated there as well.

PR #336, opened 2025-03-29
Flip align to reef face 180 degrees

Rongrrz:
# Why are we doing this?
Algae handling will require us to flip 180 degrees

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
@@ -16,7 +16,7 @@ import javax.inject.Inject;
 import static edu.wpi.first.units.Units.Degrees;
 import static edu.wpi.first.units.Units.Radians;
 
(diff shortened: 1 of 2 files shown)


src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -46,7 +46,7 @@ public void initialize() {
 
         Pose2d closestPose = pose.getClosestReefFacePose();
         idealFinalPosition = closestPose.getTranslation();
-        idealFinalHeading = closestPose.getRotation().getMeasure();
+        idealFinalHeading = closestPose.getRotation().getMeasure().plus(Degrees.of(180));
```
aschokking, line 49: ⭐ consider a comment here about why the 180, or maybe there should be a param saying if we're backwards or not that logics into this?

aschokking (approved):

PR #338, opened 2025-03-29
Potentially risky but... first generation laser can failure detection

Rongrrz:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
+++ b/src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
@@ -91,6 +91,12 @@ public class ElevatorSubsystem extends BaseSetpointSubsystem<Distance> {
     public final XDigitalInput bottomSensor;
     public final XLaserCAN distanceSensor;
 
+    public final DoubleProperty laserCanOffsetThreshold;
+    private boolean useDistanceSensor = false;
+    public int badMotorLaserCanHarmonyCount = 0;
+    public int harmonyThreshold = 3;
+    private Distance previousLaserCanDistance;
+
     private final SysIdRoutine sysId;
 
     private final ComplimentaryFilter sensorFusionFilter;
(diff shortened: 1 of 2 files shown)


src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -417,12 +424,16 @@ public void configureMotionMagicConstraints(){
         masterMotor.setTrapezoidalProfileJerk(RadiansPerSecondPerSecond.of(motionMagicJerk.get()).per(Second));
     }
 
+    public void setUseDistanceSensor(boolean value) {
```
JohnGilb, line 427: :star: for callers, it's nice when the argument is self-documenting (e.g. instead of `value`, something like `useLaserCANSensor`)

src/main/java/competition/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -440,7 +451,28 @@ public void periodic() {
...
+            if (previousLaserCanDistance == null) {
+                previousLaserCanDistance = d;
+            }
+
+            // If our laser can distance is dropping above a certain threshold, for three times
+            // All while our master motor thinks that it has positive velocity (up I am assuming) then
+            // We are in a state of not good stuff and should disable the distance sensor
+            if (previousLaserCanDistance.minus(d).in(Meters) > laserCanOffsetThreshold.get()
+                    && masterMotor.getRawVelocity().baseUnitMagnitude() > 0) {
+                if (badMotorLaserCanHarmonyCount < harmonyThreshold) {
```
JohnGilb, line 466: nit: suggest renaming to `badHarmonyBetweenLaserCanAndMotorCount` ; it's longer but makes it clear that harmony refers to the relationship between these two devices.

JohnGilb (changes requested):
For this 1st generation, just use the akitLog to record this "I want to disable the LaserCAN" value to see if this triggers at all during normal operation.

My main concern is that noise could trigger this even when nothing surprising is happening.

PR #342, opened 2025-03-29
Separate PIDManager for Align To reef algae

junyu101:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
@@ -8,6 +8,7 @@ import edu.wpi.first.math.geometry.Pose2d;
 import edu.wpi.first.math.geometry.Translation2d;
 import edu.wpi.first.units.measure.Angle;
 import xbot.common.command.BaseCommand;
+import xbot.common.math.MathUtils;
 import xbot.common.math.PIDDefaults;
 import xbot.common.math.PIDManager;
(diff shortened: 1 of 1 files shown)


src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -31,13 +36,25 @@ public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
...
+                                                 PIDManager.PIDManagerFactory pidManagerFactory) {
         this.headingModule = headingModuleFactory.create(drive.getRotateToHeadingPid());
         this.vision = vision;
         this.drive = drive;
         this.pose = pose;
         this.oi = oi;
         this.addRequirements(drive);
+
+        this.pidManager = pidManagerFactory.create("AlgaePIDManager", new PIDDefaults(
+                1.08, // P
```
aschokking, line 49: are these values the same as the current drive pid as a default?

aschokking (approved):

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
Rongrrz: These are the default (same as positional pid), I don't think we have tuned it

PR #347, opened 2025-03-30
Drive to barge

Kobeeeef:
# Why are we doing this?

Drive to barge and fix coral station drive + path

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

src/main/java/competition/subsystems/drive/commands/vision_path/PathDriveToBargeCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    public void initialize() {
+        if(coprocessor.isBargePathConfident(pose)) {
+            XTableValues.BezierCurves path = coprocessor.getLastBargePath();
+            super.setOverriddenPath(path);
+            super.setOptions(path.getOptions());
+            super.initialize();
+            super.logic.setPrioritizeRotationIfCloseToGoal(true);
+        } else {
+            log.warn("Barge path could not be determined. Cancelling.");
+            cancel();
```
aschokking, line 39: ⭐ Note: turns out `cancel()` doesn't work for commands in CommandGroups, only top level scheduled commands. So might want to have a state flag or something in `PathDriveToLocationCommand` super class that says "GaveUp" so that the parent class doesn't keep trying to do stuff in this state (though maybe it already does that if super.init isn't called?

aschokking (approved):
Next time please consider including a video in your PR description of the change working in sim to help reviewers see what it does (and that it actually works)

PR #345, opened 2025-03-30
 AlignToAlgae PID

kujo27:
# Why are we doing this?
Better PID AlignToAlgae
Asana task URL:

# Whats changing?

# Questions/notes for reviewers

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

## Video/screenshots (from simulator or live robot)

-----

PR feedback legend

 
| Symbol | Meaning                  |
|--------|--------------------------|
| :star: :star: :star:     | must be addressed                 |
| :star: :star:     | should be addressed        |
| :star:     | something to consider, a good idea                  |

PR #348, opened 2025-03-31
Algae Alignment State Machine

Rongrrz:
# Why are we doing this?
So we won't ram into the wall & can possibly steal algae from opposition

Asana task URL:

# Whats changing?
A whole new state machine! (Or am I just overcomplicating everything?)

Probably requires tuning the double properties...

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
@@ -7,6 +7,7 @@ import competition.subsystems.vision.AprilTagVisionSubsystemExtended;
 import edu.wpi.first.math.geometry.Pose2d;
 import edu.wpi.first.math.geometry.Translation2d;
 import edu.wpi.first.units.measure.Angle;
+import edu.wpi.first.wpilibj.DriverStation;
 import xbot.common.command.BaseCommand;
 import xbot.common.math.MathUtils;
 import xbot.common.math.PIDDefaults;
@@ -17,6 +18,7 @@ import xbot.common.subsystems.drive.control_logic.HeadingModule;
 import javax.inject.Inject;
 
 import static edu.wpi.first.units.Units.Degrees;
+import static edu.wpi.first.units.Units.Meters;
 import static edu.wpi.first.units.Units.Radians;
 
 public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
@@ -84,7 +86,15 @@ public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
     public void initialize() {
         log.info("Initializing");
 
-        Pose2d closestPose = pose.getClosestReefFacePose();
+        // Firstly, figure out if we are on red/blue side of the field, and we'll center accordingly
+        Pose2d currentPose = pose.getCurrentPose2d();
+
+        DriverStation.Alliance alliance = DriverStation.Alliance.Blue;
+        if (currentPose.getTranslation().getX() > PoseSubsystem.fieldXMidpointInMeters.in(Meters)) {
+            alliance = DriverStation.Alliance.Red;
+        }
+
+        Pose2d closestPose = pose.getClosestReefFacePoseByAlliance(alliance);
         idealFinalPosition = closestPose.getTranslation();
         idealFinalHeading = closestPose.getRotation().getMeasure().plus(Degrees.of(180));
     }
--- a/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
@@ -8,6 +8,7 @@ import edu.wpi.first.math.geometry.Pose2d;
 import edu.wpi.first.math.geometry.Translation2d;
 import edu.wpi.first.units.measure.Angle;
 import xbot.common.command.BaseCommand;
+import xbot.common.math.MathUtils;
 import xbot.common.math.XYPair;
 import xbot.common.subsystems.drive.control_logic.HeadingModule;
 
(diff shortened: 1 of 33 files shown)


aschokking:
Could you attach a video of it working in the sim?

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -24,33 +34,155 @@ public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
     final DriveSubsystem drive;
     final PoseSubsystem pose;
     final OperatorInterface oi;
+    final ReefCoordinateGenerator reefCoordinateGenerator;
 
     Translation2d idealFinalPosition;
     Angle idealFinalHeading;
 
+    private final PIDManager driveToAlgaePidManager;
+    private final PIDManager snapToAlgaePIDManager;
```
aschokking, line 43: This is now the 3rd open PR that adds a new PIDManager to this class? https://github.com/Team488/TeamXbot2025/pull/345

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -24,33 +34,155 @@ public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
...
+
+    final DoubleProperty interstitialThresholdToRailDrive;
+    final DoubleProperty interstitialDistance;
```
aschokking, line 46: ⭐ ⭐ consider making these DistanceProperties instead so you can't mess up the units

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -24,33 +34,155 @@ public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
...
+    final DoubleProperty interstitialApproachSpeedFactor;
```
aschokking, line 47: This can stay a DoubleProperty

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -24,33 +34,155 @@ public class AlignToNearestReefFaceForAlgaeCommand extends BaseCommand {
...
     public void initialize() {
         log.info("Initializing");
 
-        Pose2d closestPose = pose.getClosestReefFacePose();
+        // Firstly, figure out if we are on red/blue side of the field, and we'll center accordingly
+        Pose2d currentPose = pose.getCurrentPose2d();
+
+        DriverStation.Alliance alliance = DriverStation.Alliance.Blue;
+        if (currentPose.getTranslation().getX() > PoseSubsystem.fieldXMidpointInMeters.in(Meters)) {
+            alliance = DriverStation.Alliance.Red;
```
aschokking, line 119: ⭐ consider a comment about why we're checking alliance

src/main/java/competition/subsystems/pose/PoseSubsystem.java
```diff
@@ -398,16 +399,32 @@ public void ingestSimulatedSwerveModulePositions(SwerveModulePosition[] position
         this.simulatedModulePositions = Optional.of(positions);
     }
 
+    public static List<Pose2d> getReefFacePoses() {
```
aschokking, line 402: ⭐ ⭐ ⭐    Why is this one only returning the blue ones whereas getClosestReefFacePose() does red and blue?

If intentional, consider renaming this to `getBlueReefFacePoses` so it's more obvious this isn't ALL reef poses

src/main/java/competition/subsystems/vision/AprilTagVisionSubsystemExtended.java
```diff
@@ -96,6 +96,19 @@ public int getTargetAprilTagID(Landmarks.ReefFace reefFace) {
         return getTargetAprilTagID(targetReefFacePose);
     }
 
+    public int getClosestTagIdFromTranslation(Translation2d translation) {
```
aschokking, line 99: ⭐ ⭐ ⭐ this name sounds like it would be all tags, but this is only doing reef tags right?

aschokking (changes requested):

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
Rongrrz: Built on top of John Kuang's PR 😅😂

aschokking: Then that should be the base branch for your PR so I only have to review the net-new code of yours

src/main/java/competition/subsystems/drive/commands/AlignToNearestReefFaceForAlgaeCommand.java
```diff
@@ -84,4 +220,17 @@ public void execute() {
                 new XYPair()
         );
     }
+
+    public Translation2d getPowerForRelativePositionChange(Translation2d goalPosition) {
```
JohnGilb: :star: :star: Update the  BaseDriveSubsystem so it can do what you need (e.g. use a specific PIDManager) and then use that =]

src/main/java/competition/subsystems/pose/PoseSubsystem.java
```diff
@@ -398,16 +397,32 @@ public void ingestSimulatedSwerveModulePositions(SwerveModulePosition[] position
         this.simulatedModulePositions = Optional.of(positions);
     }
 
+    public static List<Pose2d> getBlueReefFacePoses() {
```
JohnGilb, line 400: :star: I know you're just extending an existing pattern, but the approach here to finding the "nearest" involves creating a bunch of objects and then disposing of them afterwards. Would be nice to have something more like the following:

1) At robot boot time, create a blue list and red list (only once)
2) At execution time, choose a given list based on what alliance you are and do the `nearest` calculation.

JohnGilb: I wouldn't change this now, but in a follow-up PR

JohnGilb (approved):

src/main/java/competition/subsystems/pose/PoseSubsystem.java
Rongrrz: Onto it!

PR #349, opened 2025-04-01
Update CoprocessorCommunicationSubsystem.java

Rongrrz:
# Why are we doing this?
Kobe forgot to revert after testing

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

Kobeeeef (approved):

aschokking (approved):
How can we avoid breaking this again in the future?

Rongrrz:
> How can we avoid breaking this again in the future?

Probably an if-else to detect if we are simulating

aschokking:
> > How can we avoid breaking this again in the future?
> 
> Probably an if-else to detect if we are simulating

Can you add it or make a task so we don't forget?

PR #357, opened 2025-04-02
Fix drive to barge & fix coprocessor

Kobeeeef:
# Why are we doing this?
[Screencast from 2025-04-02 00-57-24.webm](https://github.com/user-attachments/assets/1a1a5928-4092-45c9-ad29-fb6b50694c73)

Asana task URL:

# Whats changing?
drive to barge 

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/subsystems/drive/commands/DriveToBargeCommand.java
+++ b/src/main/java/competition/subsystems/drive/commands/DriveToBargeCommand.java
@@ -54,19 +54,23 @@ public class DriveToBargeCommand extends SwerveSimpleTrajectoryCommand {
 
     public Pose2d getNearestPoint(
             double startY, DriverStation.Alliance alliance) {
-        // Snap startY to the nearest boundary if it falls between the red and blue
-        // minimum values.
-        if (redRobotMin.get() < startY && startY < blueRobotMin.get()) {
-            if (Math.abs(startY - redRobotMin.get())
-                    < Math.abs(blueRobotMin.get() - startY)) {
-                startY = redRobotMin.get() + electricalContract.getRadiusOfRobot().div(2).in(Meters);
-            } else {
-                startY = blueRobotMin.get() + electricalContract.getRadiusOfRobot().div(2).in(Meters);
-            }
-        }
+
 
         // Determine if the alliance is BLUE.
         boolean isBlue = alliance == DriverStation.Alliance.Blue;
+
+        // Snap startY to the nearest boundary if it falls between the red and blue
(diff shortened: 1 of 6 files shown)


SeriouslyCommonLib
aschokking, line 1: ⭐ ⭐ ⭐ where's the SCL PR for these changes? Otherwise they'll get lost the next time anyone updates SCL

src/main/java/competition/subsystems/drive/commands/DriveToBargeCommand.java
```diff
@@ -53,51 +53,56 @@ public DriveToBargeCommand(BaseSwerveDriveSubsystem drive, PoseSubsystem pose,
...
+        // Snap startY to the nearest boundary if it falls between the red and blue
+        // minimum values.
+        double radius = electricalContract.getRadiusOfRobot().div(2).in(Meters);
+        if (isBlue) {
+            startY = Math.max(startY, blueRobotMin.get() + radius);
+        } else {
+            startY = Math.min(startY, redRobotMin.get() - radius);
+        }
 
         // Set heading to 180 if blue, otherwise 0.
```
aschokking, line 71: ⭐ ⭐ When we're on the opposing side of the field, shouldn't we be scoring on the barge from that side?

aschokking (changes requested):

src/main/java/competition/subsystems/vision/CoprocessorCommunicationSubsystem.java
```diff
@@ -97,8 +98,15 @@ public CoprocessorCommunicationSubsystem(
                         lastBargePath = curves;
                     }
                 }));
-        this.orinVisionCoprocessorCommander =
-                new VisionCoprocessorCommander(VisionCoprocessor.ORIN3_STATIC);
+
+        // If current instance is a simulation then use local host (if you are running the orin code locally!)
```
aschokking, line 102: 👍🏻

SeriouslyCommonLib
aschokking: Ah sorry I missed that Stephen already approved it and it was merged

aschokking (approved):
I think we should be scoring on the opponent side of the field when we're there, but that could be added in after these changes too

Kobeeeef:
> I think we should be scoring on the opponent side of the field when we're there, but that could be added in after these changes too

This already does

PR #368, opened 2025-04-04
Fix back camera

Kobeeeef:
# Why are we doing this?

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/electrical_contract/Contract2025.java
+++ b/src/main/java/competition/electrical_contract/Contract2025.java
@@ -332,8 +332,8 @@ public class Contract2025 extends ElectricalContract {
                 new CameraInfo("Apriltag_Back_Camera",
                         "AprilTagBack",
                         new Transform3d(new Translation3d(
-                                0.25 / PoseSubsystem.INCHES_IN_A_METER,
-                                -0.55,
+                                -0.55 / PoseSubsystem.INCHES_IN_A_METER,
+                                -0.25 / PoseSubsystem.INCHES_IN_A_METER,
                                 6.3 / PoseSubsystem.INCHES_IN_A_METER),
                                 new Rotation3d(0, Math.toRadians(-14.5), Math.PI)),
                         EnumSet.of(CameraCapabilities.APRIL_TAG))


src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -332,8 +332,8 @@ public CameraInfo[] getCameraInfo() {
                 new CameraInfo("Apriltag_Back_Camera",
                         "AprilTagBack",
                         new Transform3d(new Translation3d(
-                                0.25 / PoseSubsystem.INCHES_IN_A_METER,
-                                -0.55,
+                                -0.55 / PoseSubsystem.INCHES_IN_A_METER,
+                                -0.25 / PoseSubsystem.INCHES_IN_A_METER,
```
aschokking, line 336: are we sure about the sign of this?

aschokking (approved):

src/main/java/competition/electrical_contract/Contract2025.java
Kobeeeef: yes

src/main/java/competition/electrical_contract/Contract2025.java
```diff
@@ -332,8 +332,8 @@ public CameraInfo[] getCameraInfo() {
                 new CameraInfo("Apriltag_Back_Camera",
                         "AprilTagBack",
                         new Transform3d(new Translation3d(
-                                0.25 / PoseSubsystem.INCHES_IN_A_METER,
-                                -0.55,
+                                -0.55 / PoseSubsystem.INCHES_IN_A_METER,
```
aschokking, line 335: ⭐ ⭐ ⭐ actually there's no way this number is right if it's from the center of the robot, this is like 21 inches and the robot is only like 32 inches across

aschokking (changes requested):

aschokking (approved):

PR #369, opened 2025-04-04
Made switch to deadwheels automatic.

rokadias:
# Why are we doing this?
Less driver interaction needed.

Asana task URL:

# Whats changing?

# How this was tested
- [ ] tested on robot
- [x] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 
### DIFF
--- a/src/main/java/competition/Robot.java
+++ b/src/main/java/competition/Robot.java
@@ -195,6 +195,14 @@ public class Robot extends BaseRobot {
         if (!coralArmSubsystem.isCalibrated()) {
             coralArmSubsystem.forceCalibrationAtAutonomous();
         }
+        poseSub.setUseDeadwheelAssisstedPose(false);
+    }
+
+    @Override
+    public void teleopInit() {
+        super.teleopInit();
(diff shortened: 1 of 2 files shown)


src/main/java/competition/Robot.java
```diff
@@ -195,6 +195,14 @@ public void autonomousInit() {
         if (!coralArmSubsystem.isCalibrated()) {
             coralArmSubsystem.forceCalibrationAtAutonomous();
         }
+        poseSub.setUseDeadwheelAssisstedPose(false);
```
aschokking, line 198: Let's do this before 187 where we're teleporting the robot to it's starting position? 

Actually maybe it should just happen on robot boot?

aschokking (approved):
We should try it in the pit on blocks if nothing else just in case

src/main/java/competition/Robot.java
rokadias: Shouldn't make a difference for the current code, but in case someone changes the code later to be based on it.

I made both changes. Though, I expect it to be removed shortly.

PR #372, opened 2025-04-04
Clean up selection of deadwheel vs swerve poses

stephenjust:
# Why are we doing this?
We're seeing some weird jumpiness in the pose estimators. I've tried to reason through the Pose Subsystem to make sure that the pose we care about is always updated properly.

# Whats changing?

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

aschokking (approved):

rokadias (approved):
Only questionable thing is the update via wheelpositions as its untested persay.
LGTM

PR #381, opened 2025-07-30
chore: remove useless commands for seafair

NormalDuck:
# Why are we doing this?
to remove commands that are useless or potentially dangerous when there isn't a full field set up

Asana task URL: N/A

# Whats changing?

I didn't test anything since its commenting code out and I don't have a gamepad so I cannot fully test it in the sim

# How this was tested
- [ ] tested on robot
- [ ] tested in simulator
- [ ] unit tests added

-----

PR feedback legend

 

aschokking (approved):
This looks like a good set of commands to remove, thanks!

Next, given the remaining commands, are the current mappings intuitive? Would you suggest any changes to make it easier to quickly explain to someone coming up to try it?

NormalDuck:
Seafair has ended, there is no point leaving this pr open.

